PR #94, opened 2020-08-31
 Enable X-Ray tracing on FHIR API

carvantes:
Issue #, if available:

Description of changes:
Add the right permissions and enable xray tracing for the FHIR API

Screen Shot 2020-08-27 at 19 38 32

Checklist:

This PR can only be merged after the changes in https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/12 are published to npm. Then we need to bump the package version for `fhir-works-on-aws-persistence-ddb` and update the `yarn.lock` file

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -79,11 +79,19 @@ provider:
       Resource:
         - !GetAtt FHIRBinaryBucket.Arn
         - !Join ['', [!GetAtt FHIRBinaryBucket.Arn, '/*']]
+    - Action:
+        - 'xray:PutTraceSegments'
+        - 'xray:PutTelemetryRecords'
+      Effect: Allow
+      Resource:
+        - '*'
   variableSyntax: "\\${((?!AWS)[ ~:a-zA-Z0-9._@'\",\\-\\/\\(\\)]+?)}" # Use this for allowing CloudFormation Pseudo-Parameters in your serverless.yml
   logs:
     restApi:
       executionLogging: true
       fullExecutionData: true
+  tracing:
+    apiGateway: true
 
 functions:
   fhirServer:
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "fhir-works-on-aws-interface": "^1.0.0",
-    "fhir-works-on-aws-persistence-ddb": "^1.0.0",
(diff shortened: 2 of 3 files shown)


serverless.yaml
```diff
@@ -79,11 +79,19 @@ provider:
...
+      Effect: Allow
+      Resource:
+        - '*'
   variableSyntax: "\\${((?!AWS)[ ~:a-zA-Z0-9._@'\",\\-\\/\\(\\)]+?)}" # Use this for allowing CloudFormation Pseudo-Parameters in your serverless.yml
   logs:
     restApi:
       executionLogging: true
       fullExecutionData: true
+  tracing:
+    apiGateway: true
```
nguyen102, line 94: In this [blog post](https://www.serverless.com/blog/framework-release-v141) it says "If you’re using X-Ray Tracing for API Gateway you might want to enable it for your Lambda functions as well:". Should we enable lambda tracing as well? What's the difference between that and just API Gateway tracing?

Also "Due to CloudFormation limitations it's not possible to enable AWS X-Ray Tracing on existing deployments which don’t use tracing right now.

Please remove the old API Gateway and re-deploy it with tracing enabled if you want to use AWS X-Ray Tracing for API Gateway."

A headsup, that we might have to do some manual steps on the test env in the `chamonix-common` account, to allow the github workflow to successfully deploy to the test env.

carvantes: I think that only enabling tracing for APIGW is the way to go. 

The difference is that with this setup there won't be xray traces if you invoke the lambda function directly. In our case this is not an issue since the lambda function is always invoked by APIGW. APIGW has a native integration with Lambda, and enabling xray on APIGW makes the downstream Lambda calls also record xray traces

Also, I don't think Cloudformation deployment is an issue. I deployed it on my AWS account (as an update to my old stack) and it went from xray disabled to xray enabled just fine. I'd guess that the blog post is outdated. Serverless v1.41 came out in 2019-04-23. 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-tracingenabled

nguyen102 (dismissed):

rsmayda (approved):

PR #97, opened 2020-09-08
Deployment uses new RBACHandler

nguyen102:
Issue #, if available:

Description of changes:
* Moving `SUPPORTED_R4_RESOURCES` and `SUPPORTED_STU3_RESOURCES` to interface package. I'm moving the constants there because the `auth` package need the list of resources to check that a user can initiate an export request. A user needs permission to access all resources before they can initiate a system level export.


src/config.ts
```diff
@@ -78,6 +77,8 @@ export const fhirConfig: FhirConfig = {
             },
         },
     },
+
+    defaultRetryRequestInSeconds: 60 * 15,
```
carvantes: why do we need this? I don't think that customers want to customize this value. 

I'd rather start with a static non-configurable value. Also, later on when we improve the rate-limiting strategy this value can be computed by the server based on the current load of export requests.

carvantes (changes requested):

src/config.ts
nguyen102: It wasn't too hard to make this value configurable and depending on what the user picks for their persistence layer, I could foresee them wanting to change this value. They might want the retry value to be longer or shorter than 15 minutes.

carvantes: I think we should be more careful about adding config values. It seems easy, but they are part of the interface so once added we cannot take them back. 

I still don't see the case for customers wanting to customize this value. We can add it later if it is a feature request.

It's worth noting that the `Retry-After` header in 429 responses is just guidance. The client is not forced to honor it and can actually retry at whatever interval he chooses. So it's very  unlikely to be a blocker for any customer.

If we do end up having this as a config value, I don't think it should be a top level key. The name `defaultRetryRequestInSeconds` is too generic. We'd need to think better about the hierarchy of this kind of system config values. (Maybe `system.export.defaultRetryAfterSeconds` ?)

rsmayda: My 2 cents here is export is effectively some sort of operation that a customer may not want to support so I think it fits better in the `profile` section (similar to bundles)

to Nestor's other point of 'is it worth it' from a config perspective? I can see it both ways it is a trivial part of the spec that the server doesn't even have to recognize BUT it is a part of the spec. I like Nestor's point of we can always add it in but we cannot remove it. my vote is to remove it for now in favor of a constant

nguyen102: That's a good point about not being able to remove the config once it's been added. I'll remove it for now.

src/RBACRules.ts
```diff
@@ -31,7 +30,7 @@ const RBACRules: RBACConfig = {
     groupRules: {
         practitioner: {
             operations: ['create', 'read', 'update', 'delete', 'vread', 'search-type', 'transaction'],
-            resources: SUPPORTED_R4_RESOURCES,
+            resources: BASE_R4_RESOURCES,
```
rsmayda: Sadly this is an existing bug -- but if you look at the config.ts it could be STU3 making this ruleset nonsensical. Could you close that gap by just creating a constructor that either takes in the version or probably better a list of resources?

nguyen102: Updated

src/config.ts
```diff
@@ -18,7 +18,8 @@ import RBACRules from './RBACRules';
 const { IS_OFFLINE } = process.env;
 
 const fhirVersion: FhirVersion = '4.0.1';
-const authService = IS_OFFLINE ? stubs.passThroughAuthz : new RBACHandler(RBACRules, fhirVersion);
+const baseResources = fhirVersion === '4.0.1' ? BASE_R4_RESOURCES : BASE_STU3_RESOURCES;
```
rsmayda, line 21: Same as line 84 
`export const genericResources = fhirVersion === '4.0.1' ? BASE_R4_RESOURCES : BASE_STU3_RESOURCES;`

rsmayda (approved):

PR #104, opened 2020-09-15
Add export-request table

nguyen102:
Issue #, if available:

Description of changes:
* Add `export-request` table for storing export request meta data
* Change `resource` table to provisioned throughput with autoscaling. This is required for Glue jobs. [reference](https://idk.dev/how-to-export-an-amazon-dynamodb-table-to-amazon-s3-using-aws-step-functions-and-aws-glue/)
"Tables that use on-demand throughput are not currently supported." => not supported by Glue
* `export-request` table has a Global index on `jobStatus` because we need to query for `jobStatus` to figure out if we should throttle an export request.

**Related PR**
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/15
Checklist:


serverless.yaml
```diff
@@ -152,8 +167,10 @@ resources:
               KeyType: HASH
             - AttributeName: vid
               KeyType: RANGE
-          BillingMode: PAY_PER_REQUEST
           TableName: '${self:custom.resourceTableName}'
+          ProvisionedThroughput:
+            ReadCapacityUnits: 10
+            WriteCapacityUnits: 10
```
carvantes: I'd rather avoid this change unless it's absolutely required. Have you tested glue ETL with on-demand? From the AWS docs it looks like it should work

> When the DynamoDB table is in on-demand mode, AWS Glue handles the read capacity of the table as 40000. 

For exporting a large table we recommend switching your DynamoDB table to on-demand mode.
https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-connect.html#aws-glue-programming-etl-connect-dynamodb

carvantes (changes requested):

serverless.yaml
nguyen102: Ohh good find! In the tutorials that I've been following they said that DDB tables had to be in `ProvisionedThroughput` mode. But it looks like since the the tutorials were written, Glue now supports on-demand DDB tables. I just tested Glue with an on-demand DDB table, and it works.

serverless.yaml
```diff
@@ -167,6 +172,39 @@ resources:
...
+            SSEEnabled: true
+            SSEType: 'KMS'
+          BillingMode: PAY_PER_REQUEST
+          GlobalSecondaryIndexes:
+            - IndexName: '${self:custom.exportRequestTableJobStatusIndex}'
+              KeySchema:
+                - AttributeName: jobStatus
+                  KeyType: HASH
+                - AttributeName: jobOwnerId
+                  KeyType: RANGE
```
rsmayda, line 200: Not sure if this matters but I assume `ownerId` will be more unique than `jobStatus` could we flip which is primary & sort key?
```suggestion
                - AttributeName: jobOwnerId
                  KeyType: HASH
                - AttributeName: jobStatus
                  KeyType: RANGE

```

nguyen102: Ah the reason for `jobStatus` being the hashKey of the index is so we can query on `jobStatus` by itself. I think if `jobStatus` was the rangeKey, we woudn't be able to search for jobs that has `jobStatus === in-progress`

carvantes (approved):

PR #112, opened 2020-09-24
fix: Update DynamoDB sort key from String to Num

### LINKED ISSUE #108: 'ResourceNotFoundError' when updating a Resource
Good afternoon

We noticed that when updating a resource with a document status of 'AVAILABLE' and vid of 10 we were receiving a 'ResourceNotFoundError' error. The latest 2 versions of the resource should be returned from the 'getMostRecentValidResource' query, but we were actually getting versions 8 & 9 back, which were both set to 'DELETED'. The problem appears to be related to the vid, which is defined as a string in ddb and therefore affects the sort order, as a string of '10' is less than '8' & '9'. I did think about updating the vid to be a number in ddb, but the vid is passed in as a string parameter.

I just wanted to check if this was a real issue or if I was doing something wrong?

Kind regards

Steve

rsmayda:
This will allow resources >10 versions to be created.

Changes:
* Added migration script to be ran locally
* Updated dependencies

Issue #, if available: #108 

Checklist:


yarn.lock
```diff
@@ -1253,6 +1253,15 @@
     "@types/qs" "*"
```
rsmayda: Lock file to be updated once Persistence is pushed to NPM registry

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.2.0] - 2020-09-25
+
+### Added
+
+- fix: Update CFN to change Dynamo sort key
```
carvantes:
```suggestion
- fix: Update CloudFormation template to change DynamoDB table key schema

```

serverless.yaml
```diff
@@ -175,6 +181,36 @@ resources:
...
+          BillingMode: PAY_PER_REQUEST
+          TableName: '${self:custom.oldResourceTableName}'
+          SSESpecification:
+            KMSMasterKeyId: !Ref DynamodbKMSKey
+            SSEEnabled: true
+            SSEType: 'KMS'
+          Tags: !If
+            - isNotDev
+            - - Key: 'backup'
+                Value: 'daily'
```
carvantes: do we want to keep making backups of the unused table?

serverless.yaml
```diff
@@ -136,25 +137,30 @@ functions:
...
           startingPosition: LATEST
 
 resources:
+  - Description: FHIR Work on AWS Example Server; v1.2
   - Conditions:
       isDev: !Equals ['${self:custom.stage}', 'dev']
       isNotDev: !Not [Condition: isDev]
   - Resources:
-      ResourceDynamoDBTable:
+      ResourceDBTable:
```
carvantes: Maybe name it something like `ResourceDynamoDBTableV2`

I think we should use a more specific name to avoid confusion. There's no way to know which table is the good one between `ResourceDynamoDBTable` and `ResourceDBTable` other than looking at the template comments.

scripts/sort-key-migration.js
```diff
@@ -0,0 +1,101 @@
...
+    console.log(`Writing to new table: ${NEW_RESOURCE_TABLE}`);
+
+    const writeResults = await Promise.all(batchPromises);
+
+    console.log('Looking for BatchWrite Errors');
+    for (let i = 0; i < writeResults.length; i += 1) {
+        const result = writeResults.UnprocessedItems;
+        if (result && result[NEW_RESOURCE_TABLE]) {
+            console.error('Unprocessed Entry:');
+            console.error(result[NEW_RESOURCE_TABLE]);
```
carvantes: What should customers do if there are errors? 
Looks like this script is idempotent. Maybe print to console that they can safely retry the whole thing.

scripts/sort-key-migration.js
```diff
@@ -0,0 +1,101 @@
...
+ * - Install node version 12+: https://nodejs.org/en/download/
+ * - Ensure your user has SCAN access to OLD_TABLE and WRITE access to NEW_TABLE
+ * Required path params for this script:
+ * - REGION - AWS Region both DBs are in
+ * - ACCESS_KEY - AWS Access Key
+ * - SECRET_KEY - AWS Secret Key
+ * - OLD_TABLE - The old table's name. It should be: ('resource-<stage>')
+ * - NEW_TABLE - The new table's name. It should be: ('resource-db-<stage>')
+ * Usage:
+ * - Command: node -e 'require("./sort-key-migration").handler()'
```
carvantes: can we refactor to just run this with `node ./sort-key-migration.js`?

carvantes:
nit: You should sync your local repo before submitting PRs to avoid adding 40+ redundant commits

scripts/sort-key-migration.js
```diff
@@ -0,0 +1,117 @@
...
+            console.log('End of current scan; starting another');
+            scanParam.ExclusiveStartKey = scanResult.LastEvaluatedKey;
+        }
+    } while (scanResult.LastEvaluatedKey);
+    if (batchWrites.length > 0) {
+        console.log(`Batch write size: ${batchWrites.length}`);
+        batchPromises.push(DynamoDb.batchWriteItem(createBatchWriteRequest(batchWrites)).promise());
+        batchWrites = [];
+    }
+    console.log(`Writing to new table: ${NEW_RESOURCE_TABLE}`);
```
carvantes: fyi at this points most entries have already been written to DDB. The write request is sent the moment `DynamoDb.batchWriteItem()` is called. The end result is correct though.

serverless.yaml
```diff
@@ -136,25 +137,27 @@ functions:
...
             'Fn::GetAtt':
-              - ResourceDynamoDBTable
+              - ResourceDynamoDBTableV2
               - StreamArn
           batchSize: 100
           maximumRetryAttempts: 3
           startingPosition: LATEST
 
 resources:
+  - Description: FHIR Work on AWS Example Server; v1.2
```
carvantes: It's kind of nice to display the version somewhere, But It's very likely that we'll forget to update this description for future releases. 

We need a mechanism to ensure this is kept in sync. Or just remove the version until we have one.

Also, shouldn't it be v2.0.0?

carvantes (changes requested):

serverless.yaml
rsmayda: Good call removing for now

carvantes (dismissed):

carvantes (approved):

PR #114, opened 2020-10-01
feat: Deploying FWoA with Integration Transform

nguyen102:
**Story**
FHIR-370

**Branching** 
I created an `api` branch that we can merge Integration Transform related code changes to. It'll allow us to have a branch that can be used to deploy a FHIR Server with an Integration Transform persistent layer.

**Description of changes:**
* In this PR we will handle spinning up FWoA with Integration Transform. The code works locally by pulling in `persistence-ddb` package with the branch `api`. Once we have permission to publish `persistence-api` repo, we will change the dependency in `deployment` to use `persistence-api`.
* We add `AWS SSM` to allow us to grab the Integration Transform url from `AWS Parameter Store`
* We no longer use `ElasticSearch`, `DDB`, and `Binary S3 Bucket` so those have been removed
* Added a `SecretsManager` class to handle grabbing secrets from `AWS Param Store`

**Next PR**
* In the next PR, we will add documentation to `deployment`. The documentation will detail what needs to be set up to run this code. That requirement is captured in this ticket (FHIR-371)

**Checklist:**


yarn.lock
```diff
@@ -4004,59 +3904,57 @@ fecha@^4.2.0:
...
+"fhir-works-on-aws-search-es@file:.yalc/fhir-works-on-aws-search-es":
   version "1.1.0"
-  resolved "https://registry.yarnpkg.com/fhir-works-on-aws-search-es/-/fhir-works-on-aws-search-es-1.1.0.tgz#b0856dc1990ef277a5a8359b25556d6882d9b303"
-  integrity sha512-kvprngWZg85P0vXOlQcakbtigrQEBC3bVPAbEdXVBuKqzMYpXd8ycuuMiDiFi/G9iV20Y5/Pu1cX9KBRHS7SHA==
   dependencies:
     "@elastic/elasticsearch" "7"
     aws-elasticsearch-connector "^8.2.0"
     aws-sdk "^2.610.0"
-    fhir-works-on-aws-interface "^2.0.0"
+    fhir-works-on-aws-interface "file:../../../../Users/thingut/Library/Caches/Yarn/v6/npm-fhir-works-on-aws-search-es-1.1.0-530d5c7a-196d-45f8-88c6-72c947dafe80-1601912260014/node_modules/fhir-works-on-aws-search-es/.yalc/fhir-works-on-aws-interface"
```
carvantes: Local file references should not be merged

src/secretsManager.ts
```diff
@@ -0,0 +1,36 @@
+import AWS from 'aws-sdk';
+
+const { IS_OFFLINE, AWS_REGION, STAGE } = process.env;
+export default class SecretsManager {
```
carvantes: Why is this called `SecretsManager`?

AWS.SSM is AWS Systems Manager
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SSM.html

src/index.ts
```diff
@@ -5,14 +5,14 @@
...
-export default async (event: any = {}, context: any = {}): Promise<any> => {
-    return serverlessHandler(event, context);
+export default async (event1: any = {}, context: any = {}): Promise<any> => {
+    const config = await getConfig();
+    const serverlessHandler = serverless(generateServerlessRouter(config, genericResources), {
+        request(request: any, event2: any) {
+            request.user = event2.user;
+        },
+    });
+    return serverlessHandler(event1, context);
```
carvantes: This is adding ~600ms of latency to every request since all requests are building the entire `serverlessHandler` object.

The `serverlessHandler` should only be built once(on cold starts) and then be reused by all subsequent requests. 

You can do something like:
```js
const serverlessHandler: Promise<ServerlessHttp.Handler> = asyncFnThatCallsGetConfigAndBuildsTheServerlessHandler()

export default async (event: any = {}, context: any = {}): Promise<any> => {
    return (await serverlessHandler)(event, context);
};
```

carvantes (changes requested):

src/secretsManager.ts
```diff
@@ -0,0 +1,36 @@
...
+                region: 'us-west-2',
+                accessKeyId: process.env.ACCESS_KEY,
+                secretAccessKey: process.env.SECRET_KEY,
+            });
+        }
+    }
+
+    static async getIntegrationTransformUrl(): Promise<string> {
+        const ssm = new AWS.SSM();
+        let path = process.env.INTEGRATION_TRANSFORM_PATH ?? 'fake_path';
```
carvantes: Avoid assigning values such as `fake_path`

src/secretsManager.ts
```diff
@@ -0,0 +1,36 @@
...
+                Name: path,
+                WithDecryption: true,
+            })
+            .promise();
+
+        if (data.Parameter?.Value === undefined) {
+            console.error(`Unable to get INTEGRATION_TRANSFORM_PATH. Path: ${path}`);
+        }
+
+        return data.Parameter?.Value ?? '';
```
carvantes: Why are we returning an empty string? 

I think it's better to fail explicitly. Otherwise you'd only know about the error much later, once a persistence api call is made.

src/secretsManager.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+SecretsManager.initialize();
```
carvantes: It's weird to have to call a static class method before exporting. Do consumers of this class have any reason to call `initialize `?

I think that this file should simply export a function like `getIntegrationTransformUrl `. I don't think a class makes sense here if it's only going to have public static methods.

src/secretsManager.ts
```diff
@@ -0,0 +1,36 @@
+import AWS from 'aws-sdk';
```
carvantes: Add the Copyright License Header on this file

carvantes (changes requested):

src/index.ts
carvantes: btw, this requires a bit more sophisticated error handling, since if the serverlessHandler fails to build for a transient error (i.e. network error when calling SSM) you don't want to "cache" that and have all subsequent requests fail.

But the above is the basic idea

yarn.lock
nguyen102: good find. I'll delete that.

src/secretsManager.ts
nguyen102: Yeah, I was debating what to put here. My thought was to either use an empty string or `fake_path`. I decided on `fake_path` because I felt it would make the error clearer when the error is printed out on this line.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/891ccdb259e44d9e9e83f03e4a52d67f0d7af7a3/src/secretsManager.ts#L29

I'm open to other ways of handling it though.

src/secretsManager.ts
nguyen102: added

src/secretsManager.ts
nguyen102: In this `package`, we only make calls to AWS in this class. Therefore, `initialize` only needs to be run when this class is being used. 

I wanted to encapsulate the logic of `initialize` and making requests to System Manager param store in one place. The alternative would be to put the `initialize` code in `config.ts` also and export `getIntegrationTransformUrl`. However, none of the other code in `config` needs to make calls to AWS.

src/secretsManager.ts
nguyen102: Yeah, initially I was thinking about throwing an error here instead of returning an empty string. However, upon testing, it produced a confusing user experience.

When a request is made, the `FhirConfig` is initialized first, and then the routing layer is initialized. If we throw an error here, while the `FhirConfig` is being set up, the routing layer won't be initialized yet.

In testing this results in every request to the FHIR Server returning with a 200 response and an empty body.

src/secretsManager.ts
nguyen102: Good question. I named it `SecretsManager` to describe it's job instead of it's implementation. This class is responsible for fetching secrets. The implementation detail is that it uses  `AWS Systems Manager`.

carvantes: Still, that name is very misleading. Also, what secrets are we fetching? The integration transform url is not a secret

src/index.ts
nguyen102: After discussing offline, we can throw errors if we can't retrieve the Integration Transform URL from SSM. Lambda will automatically terminate the container. We won't have to worry about lambda "caching" the wrong value.
https://serverlessfirst.com/fff-function-initialisation/

src/secretsManager.ts
nguyen102: Throwing the error here will allow lambda to terminate the error and return a 502 error. This code would be used during initialization of the FhirConfig

src/secretsManager.ts
nguyen102: Once deployed to lambda, throwing errors will return a 502 response. It seems only `serverless-offline` will return a 200. Therefore, I've shifted to throwing errors in this class.

src/secretsManager.ts
nguyen102: Updated to `ParameterStore`.

src/secretsManager.ts
nguyen102: Updated code to have this file export functions while still keeping updates of AWS config in this file.

src/parameterStore.ts
```diff
@@ -0,0 +1,43 @@
...
+        path = `fhir-service.integration-transform.${AWS_REGION}.${STAGE}.url`;
+    }
+    const data = await ssm
+        .getParameter({
+            Name: path,
+            WithDecryption: true,
+        })
+        .promise();
+
+    if (data.Parameter?.Value === undefined) {
```
carvantes: when does this happen? doesn't it throw a `ParameterNotFound` error if the param does not exist?

nguyen102: This was to deal with eslint complaining that `return data.Parameter.Value` might be null. But since that can't happen because `getParameter` will throw an error, I'll delete this if statement.

scripts/iam_policy.json
```diff
@@ -23,14 +23,24 @@
           "kms:*",
           "states:*",
           "backup:*",
-          "backup-storage:MountCapsule"
+          "backup-storage:MountCapsule",
+          "ssm:GetParameter"
```
carvantes, line 27: Not needed.

This file is for permissions needed to deploy the solution. Not for permissions that are needed at runtime

scripts/iam_policy.json
```diff
@@ -23,14 +23,24 @@
...
         "Resource": "*"
       },
       {
         "Effect": "Allow",
         "Action": ["apigateway:*"],
         "Resource": "arn:aws:apigateway:*::/*"
+      },
+      {
+        "Action": [
+          "execute-api:Invoke"
```
carvantes, line 38: Not needed.

This file is for permissions needed to deploy the solution. Not for permissions that are needed at runtime

src/parameterStore.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const data = await ssm
+        .getParameter({
+            Name: path,
+            WithDecryption: true,
+        })
+        .promise();
+
+    // @ts-ignore
+    return data.Parameter.Value;
```
carvantes:
```suggestion
    return data.Parameter!.Value!;
```
(test this before merging suggestion)
avoid ts-ignore unless absolutely necessary. Here we can explicitly assert that `Parameter.value` is not undefined. afaik that's always true. It looks like the aws-sdk just have sub-optimal type definitions

src/parameterStore.ts
```diff
@@ -0,0 +1,42 @@
...
+// eslint-disable-next-line import/prefer-default-export
+export async function getIntegrationTransformUrl(): Promise<string> {
+    const ssm = getSSM();
+    let path = process.env.INTEGRATION_TRANSFORM_PATH;
+    if (path === undefined) {
+        throw new Error('INTEGRATION_TRANSFORM_PATH is not defined in environment variables');
+    }
+    if (IS_OFFLINE === 'true') {
+        path = `fhir-service.integration-transform.${AWS_REGION}.${STAGE}.url`;
+    }
```
carvantes, line 31: Is the order of If conditions right? This looks like it will always fail in offline mode

carvantes (approved):

src/parameterStore.ts
nguyen102: `serverless offline` renders the environment variable as `[object Object]` so technically it's not rendered as `undefined`. The code still works correctly.

This code catches case in which someone edits `serverless.yaml` and deletes the environment variable for `INTEGRATION_TRANSFORM_PATH`

scripts/iam_policy.json
nguyen102: When customers want to run the code locally for development purposes, they would need to run `sls offline start` with an AWS account. The "offline" lambda that is spun up would use that AWS account to run FWoA code. That AWS account needs permission to grab parameters from SSM and to also invoke the API GW of  IntTran. 

I'll update the doc here to be more specific as to which AWS account is required when spinning up FWoA in `offline` mode.

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/DEVELOPMENT.md

scripts/iam_policy.json
nguyen102: When customers want to run the code locally for development purposes, they would need to run `sls offline start` with an AWS account. The "offline" lambda that is spun up would use that AWS account to run FWoA code. That AWS account needs permission to grab parameters from SSM and to also invoke the API GW of  IntTran. 

I'll update the doc here to be more specific as to which AWS account is required when spinning up FWoA in `offline` mode.

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/DEVELOPMENT.md

src/parameterStore.ts
nguyen102: tested and it works. Thanks for the suggestion.

PR #116, opened 2020-10-08
chore: Update docs to reflect using IntTran and resource cleanup

nguyen102:
Description of changes:
* Update docs to reflect usage with Integration Transform, including creating the proper Parameter Store values
* `ApiDataService` needs to know what region the Integration Transform is deployed in for proper AuthN with Integration Transform. [Related PR](https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/25)
*  Removed references to DDB, ES, and S3 Binary Bucket

Checklist:


serverless.yaml
```diff
@@ -14,6 +14,7 @@ custom:
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
   integrationTransformPath: '${self:service}.integration-transform.${self:custom.region}.${self:custom.stage}.url'
+  integrationTransformAwsRegionPath: '${self:service}.integration-transform.${self:custom.region}.${self:custom.stage}.awsRegion'
```
carvantes: Use `/` instead of `.` in the parameter name.
SSM has the concept of hierarchical parameters and prefixes. It helps keep params organized and makes it easier to manage permissions.

Also, maybe it's not worth including `${self:custom.region}` in the param name since params only exist in a region anyways.

https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-create.html

https://aws.amazon.com/blogs/mt/organize-parameters-by-hierarchy-tags-or-amazon-cloudwatch-events-with-amazon-ec2-systems-manager-parameter-store/

carvantes (changes requested):

src/parameterStore.ts
```diff
@@ -39,3 +38,16 @@ export async function getIntegrationTransformUrl(): Promise<string> {
...
 }
+
+export default async function getIntegrationTransformData(): Promise<{
+    integrationTransformUrl: string;
+    integrationTransformAwsRegion: string;
+}> {
+    const promises = [
+        getParam('INTEGRATION_TRANSFORM_PATH', 'url'),
+        getParam('INTEGRATION_TRANSFORM_AWS_REGION_PATH', 'awsRegion'),
+    ];
```
carvantes: We should retrieve all the parameters in a single call. use the get-parameters API instead

carvantes (changes requested):

src/parameterStore.ts
nguyen102: updated

serverless.yaml
nguyen102: Updated to `/fhir-service/integration-transform/<STAGE>/url` and `/fhir-service/integration-transform/<STAGE>/awsRegion`

scripts/iam_policy.json
```diff
@@ -24,7 +24,8 @@
           "states:*",
           "backup:*",
           "backup-storage:MountCapsule",
-          "ssm:GetParameter"
+          "ssm:GetParameter",
+          "ssm:GetParameters"
```
nguyen102: The IAM user that is created here is used to deploy FWoA to AWS. This IAM user can also be used during development to run `serverless offline`. If the IAM User is being used during development, the user should also have `ssm:GetParameters` permission.

This info is referenced here.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/api/DEVELOPMENT.md#development

src/parameterStore.ts
```diff
@@ -19,23 +19,56 @@ function getSSM() {
...
-        .getParameter({
-            Name: path,
+        .getParameters({
+            Names: paramStorePaths,
             WithDecryption: true,
         })
         .promise();
 
-    return data.Parameter!.Value!;
+    if (data.InvalidParameters === undefined || data.InvalidParameters.length > 0) {
```
carvantes: Why does `data.InvalidParameters === undefined` causes an error?

src/parameterStore.ts
```diff
@@ -19,23 +19,56 @@ function getSSM() {
...
+    integrationTransformUrl: string;
+    integrationTransformAwsRegion: string;
+}> {
+    const integrationTransformUrlPath = getParamPath(
+        'INTEGRATION_TRANSFORM_PATH',
+        `/fhir-service/integration-transform/${STAGE}/url`,
+    );
+    const integrationTransformAwsRegionPath = getParamPath(
+        'INTEGRATION_TRANSFORM_AWS_REGION_PATH',
+        `/fhir-service/integration-transform/${STAGE}/awsRegion`,
```
carvantes, line 66: What's the value of `STAGE` when running sls offline ?

scripts/iam_policy.json
carvantes: I don't agree with this. Permissions to deploy and permissions needed at runtime are a very different thing.

This file is already a bit of a mess (what does  `iam_policy.json` means? only the docs tell you that this is meant for deployment, we should probably rename it. Also, It has too many `*` permissions). I don't think that adding runtime permissions on top of it is a good idea.

When running sls offline you can either:
- Assume an Admin role. It's ok, you are a dev testing stuff in a dev AWS account. You can use an admin role to not get any permission issues and just focus on development.
- Assume a role that matches the permissions of the lambda function role that you are testing. I **think** that serverless has something to help set this up(when testing individual lambda functions). Running it this way is closer to how the lambda runs once deployed on AWS and can be helpful to test permissions.

src/parameterStore.ts
nguyen102: `eslint` tagged that `InvalidParameters` could be undefined and this was an attempt to fix the `eslint` issue. But actually using `!` is a better way to handle it. 

I'm going to change it to `if (data.InvalidParameters!.length > 0) {`

src/parameterStore.ts
nguyen102: By default `stage` is set to be `dev`. But it's configurable, you can run `sls offine start --stage <STAGE>`

src/parameterStore.ts
carvantes: That'd be better. Although in that case you need to make sure that the response always includes `InvalidParameters`(even if it is an empty array)

btw Typescript is the one that highlights this error. eslint doesn't care

scripts/iam_policy.json
nguyen102: I like the idea of AdminitrativeAccess. That's actually how I currently run the code locally too. I'll revert the changes here, and update the text in DEVELOPMENT.md

src/parameterStore.ts
nguyen102: Looking at the [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SSM.html#getParameters-property). It looks like `InvalidParameters` is always present, but is an empty array when there are no `invalid` parameters. Also when I tested, I saw that there was an empty array for `InvalidParameters`

carvantes: Sounds good. I think I've seen similar situations with other aws response types. it seems that the Type definitions for the aws-sdk are not that great...

INSTALL.md
```diff
@@ -302,96 +302,23 @@ AWS_ACCESS_KEY_ID=<ACCESS_KEY> AWS_SECRET_ACCESS_KEY=<SECRET_KEY> python3 script
...
-#### DynamoDB table backups
-
-Daily DynamoDB Table back-ups can be optionally deployed via an additional 'fhir-server-backups' stack. The installation script will deploy this stack automatically if indicated during installation.
+aws ssm put-parameter --region $region --cli-input-json \
+'{"Type": "String", "Name": "/fhir-service/integration-transform/<STAGE>/awsRegion", "Value": "<AWS-REGION>"}'
 
-The reason behind multiple stacks is that backup vaults can be deleted only if they are empty, and you can't delete a stack that includes backup vaults if they contain any recovery points. With separate stacks it is easier for you to operate.
 
-These back-ups work by using tags. In the [serverless.yaml](./serverless.yaml) you can see ResourceDynamoDBTableV2 has a `backup - daily` & `service - fhir` tag. Anything with these tags will be backed-up daily at 5:00 UTC.
+Exp
```
carvantes: Exp?

carvantes (approved):

INSTALL.md
nguyen102: It's supposed to be an abbreviation for example, I'll change it to be `E.g.` to be clearer what it means

PR #121, opened 2020-10-10
feat: CloudFormation template distribution and changes for AWS Solution package

swoldemi:
Issue #, if available: None

Description of changes:
1. Added script to package and modify a deployment CloudFormation template. The template and Lambda code artifacts are uploaded to a specified bucket (e.g. solutions-us-east-1).
2. Added CloudFormation parameters to make changes at Stack creation time possible (versus using sls variables)
3. Added AWS::Cognito::UserPoolUser for Kibana dev access. Email with random password is sent to KibanaUserEmail when stacked is created and isDev == True
4. Added AWS::IAM::User to as alternative to providing IAM user ARN in serverless_config.json
5. Other misc. template changes and resource modifications/additions

Script used for local testing:
```bash
#!/bin/bash

# Change prefix and region to use a bucket in your account
BUCKET_PREFIX=swoldemi-secondary
BUCKET_REGION=us-east-1
STACK_NAME=fhir-test-1

./scripts/s3-dist.sh $BUCKET_PREFIX $BUCKET_REGION && \
aws --region $BUCKET_REGION cloudformation create-stack --stack-name $STACK_NAME \
	--parameters 'ParameterKey=KibanaUserEmail,ParameterValue=<YOUR_EMAIL_HERE>' \
	--template-url https://$BUCKET_PREFIX-$BUCKET_REGION.s3.amazonaws.com/cloudformation-template-update-stack.json \
	--capabilities CAPABILITY_NAMED_IAM && \
rain watch --region $BUCKET_REGION $STACK_NAME
```

Template linted with [cfn-python-lint](https://github.com/aws-cloudformation/cfn-python-lint):
```
cfn-lint -t .serverless/cloudformation-template-update-stack.json
```
There are 3 linting rules can't be solved (I think) because of resources created/associated by `sls package` under the hood: https://gist.github.com/swoldemi/180b1269ed24834a35cfa9e3e649d1e9

Merging fhirServer.events.http into a single event looks to work. Tested with...
```
curl -H "Accept: application/json" -H "Authorization:$TOKEN" -H "x-api-key:$API_KEY" $GATEWAY_ENDPOINT/metadata
```
...after running `provision-user.py` for token and retrieving API key from API Gateway

Kibana dashboard access confirmed by going to ElasticSearchDomainKibanaEndpoint output, typing in provided email and password sent to email inbox, and changing password when prompted.

Checklist:

### DIFF
--- a/cloudformation/alarms.yaml
+++ b/cloudformation/alarms.yaml
@@ -5,11 +5,10 @@
 
 Resources:
   DDBToESErrorAlarm:
-    DependsOn: DdbToEsLambdaFunction
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when the Stream errors is more than 1 unit for 15 minutes out of the past 25 minutes. Streams do have retry logic
-      AlarmName: FhirSolution.${self:custom.stage}.High.DDBToESLambdaErrorAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, High, DDBToESLambdaErrorAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 5
@@ -25,11 +24,10 @@ Resources:
       TreatMissingData: notBreaching
       Unit: Count
   FhirLambdaErrorAlarm:
-    DependsOn: FhirServerLambdaFunction
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when Fhir errors is more than 1
-      AlarmName: FhirSolution.${self:custom.stage}.High.FhirLambdaErrorAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, High, FhirLambdaErrorAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 1
@@ -44,11 +42,10 @@ Resources:
       TreatMissingData: notBreaching
       Unit: Count
   FhirLambdaLatencyAlarm:
-    DependsOn: FhirServerLambdaFunction
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when Fhir average latency is more than 2.5s; 2 times
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirLambdaLatencyAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirLambdaLatencyAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 5
@@ -68,7 +65,7 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when Api GW has more than 1 5xx errors; 3 times
-      AlarmName: FhirSolution.${self:custom.stage}.High.ApiGateway5XXErrorAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, High, ApiGateway5XXErrorAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 5
@@ -76,7 +73,7 @@ Resources:
       MetricName: 5XXError
       Dimensions:
         - Name: ApiName
-          Value: ${self:custom.stage}-${self:service}
+          Value: !Join ["-", [!Ref Stage, "${self:service}"]]
       Namespace: AWS/ApiGateway
       Period: 300
       Statistic: Sum
@@ -88,7 +85,7 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when Api GW has more than 1 4xx errors; 3 times
-      AlarmName: FhirSolution.${self:custom.stage}.Low.ApiGateway4XXErrorAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, ApiGateway4XXErrorAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 5
@@ -96,7 +93,7 @@ Resources:
       MetricName: 4XXError
       Dimensions:
         - Name: ApiName
-          Value: ${self:custom.stage}-${self:service}
+          Value: !Join ["-", [!Ref Stage, "${self:service}"]]
       Namespace: AWS/ApiGateway
       Period: 300
       Statistic: Sum
@@ -108,7 +105,7 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: Alarm when Api GW average latency is more than 3s; 2 times
-      AlarmName: FhirSolution.${self:custom.stage}.Low.ApiGatewayLatencyAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, ApiGatewayLatencyAlarm]]
       ActionsEnabled: False
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 5
@@ -116,7 +113,7 @@ Resources:
       MetricName: Latency
       Dimensions:
         - Name: ApiName
-          Value: ${self:custom.stage}-${self:service}
+          Value: !Join ["-", [!Ref Stage, "${self:service}"]]
       Namespace: AWS/ApiGateway
       Period: 60
       Statistic: Average
@@ -126,7 +123,7 @@ Resources:
   ClusterStatusRedAlarm:
     Type: 'AWS::CloudWatch::Alarm'
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.High.FhirESClusterStatusRedAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, High, FhirESClusterStatusRedAlarm]]
       AlarmDescription: 'Primary and replica shards of at least one index are not allocated to nodes in a cluster.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -145,7 +142,7 @@ Resources:
     Type: 'AWS::CloudWatch::Alarm'
     Condition: isNotDev
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterStatusYellowAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterStatusYellowAlarm]]
       AlarmDescription: 'Replica shards for at least one index are not allocated to 2 nodes in a cluster.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -163,7 +160,7 @@ Resources:
   ClusterCPUUtilizationTooHighAlarm:
     Type: 'AWS::CloudWatch::Alarm'
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.High.FhirESClusterCPUUtilAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, High, FhirESClusterCPUUtilAlarm]]
       AlarmDescription: 'Average CPU utilization over last 10 minutes too high.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -181,7 +178,7 @@ Resources:
     Type: 'AWS::CloudWatch::Alarm'
     Condition: isNotDev
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterMasterCPUUtilAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterMasterCPUUtilAlarm]]
       AlarmDescription: 'Average CPU utilization over last 10 minutes too high.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -198,7 +195,7 @@ Resources:
   ClusterFreeStorageSpaceTooLowAlarm:
     Type: 'AWS::CloudWatch::Alarm'
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterFreeStorageSpaceTooLowAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterFreeStorageSpaceTooLowAlarm]]
       AlarmDescription: 'Cluster is running out of storage space.'
       ComparisonOperator: LessThanThreshold
       Dimensions:
@@ -215,7 +212,7 @@ Resources:
   ClusterIndexWritesBlockedTooHighAlarm:
     Type: 'AWS::CloudWatch::Alarm'
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterIndexWritesBlockedTooHighAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterIndexWritesBlockedTooHighAlarm]]
       AlarmDescription: 'Cluster is blocking incoming write requests.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -232,7 +229,7 @@ Resources:
   ClusterJVMMemoryPressureTooHighAlarm:
     Type: 'AWS::CloudWatch::Alarm'
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterJVMMemoryAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterJVMMemoryAlarm]]
       AlarmDescription: 'Average JVM memory pressure over last 10 minutes too high.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -250,7 +247,7 @@ Resources:
     Type: 'AWS::CloudWatch::Alarm'
     Condition: isNotDev
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterMasterJVMMemoryAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterMasterJVMMemoryAlarm]]
       AlarmDescription: 'Average JVM memory pressure over last 10 minutes too high.'
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
@@ -268,7 +265,7 @@ Resources:
     Type: 'AWS::CloudWatch::Alarm'
     Condition: isNotDev
     Properties:
-      AlarmName: FhirSolution.${self:custom.stage}.Low.FhirESClusterMasterNotReachableFromNodeAlarm
+      AlarmName: !Join [".", [FhirSolution, !Ref Stage, Low, FhirESClusterMasterNotReachableFromNodeAlarm]]
       AlarmDescription: 'Master node stopped or not reachable. Usually the result of a network connectivity issue or AWS dependency problem.'
       ComparisonOperator: LessThanThreshold
       Dimensions:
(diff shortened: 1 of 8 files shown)


cloudformation/alarms.yaml
```diff
@@ -5,11 +5,10 @@
 
 Resources:
   DDBToESErrorAlarm:
-    DependsOn: DdbToEsLambdaFunction
```
rsmayda, line 8: Why remove this?

cloudformation/elasticsearch.yaml
```diff
@@ -81,10 +99,23 @@ Resources:
...
+      UserAttributes:
+        - Name: email
+          Value: !Ref KibanaUserEmail
+      Username: !Ref KibanaUserEmail
+      UserPoolId: !Ref KibanaUserPool
+
   ElasticSearchDomain:
     Type: AWS::Elasticsearch::Domain
     Properties:
+      DomainName: fhir-domain
```
rsmayda: Why are we setting this? If we are we will need to add more dimensionality (e.g. adding stage)

scripts/s3-dist.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -e
+
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script should be run from the repo's root directory (same directory as serverless.yaml)
```
rsmayda, line 11: Add a description of what this script is

scripts/s3-dist.sh
```diff
@@ -0,0 +1,42 @@
...
+# cd fhir-works-on-aws-deployment
+# chmod +x ./scripts/s3-dist.sh && ./scripts/s3-dist.sh <BUCKET_NAME> <BUCKET_REGION>
+
+# Package output directory is assumed to be default (.serverless). Passing -p flag to `sls package` can break this script.
+# Dependencies: npm, aws CLI (configured), sed, jq
+if [ -z "$1" ] || [ -z "$2" ]; then
+    echo Please provide both the prefix and region for the bucket
+    echo For example: "$0" solutions us-east-1
```
rsmayda, line 19: What does this echo to cmd line? just confused by the `$0`

scripts/s3-dist.sh
```diff
@@ -0,0 +1,42 @@
...
+    exit 1
+fi
+
+# Bucket name is assumed to be $BUCKET_PREFIX-$BUCKET_REGION (e.g. solutions-us-east-1)
+BUCKET_PREFIX=$1
+BUCKET_REGION=$2
+
+function cleanup(){
+    echo Removing deploymentBucket from serverless.yaml
+    sed -i -e '20d' serverless.yaml
```
rsmayda: What is this doing removing line 20?

serverless.yaml
```diff
@@ -1,362 +1,418 @@
...
+  config: ${file(serverless_config.json)}
+
+provider:
+  name: aws
+  region: us-west-2
+  stage: dev
+  runtime: nodejs12.x
+  memorySize: 512
+  stackTags:
+    FHIR_SERVICE: !Sub 'fhir-service-${AWS::Region}-${self:custom.stage}'
```
rsmayda: Do you want to use !Ref Stage here as well?

serverless.yaml
```diff
@@ -1,362 +1,418 @@
...
+    S3_KMS_KEY:
+      Ref: S3KMSKey
+    RESOURCE_TABLE: '${self:custom.resourceTableName}'
+    FHIR_BINARY_BUCKET:
+      Ref: FHIRBinaryBucket
+    ELASTICSEARCH_DOMAIN_ENDPOINT: !Join ['', ['https://', !GetAtt ElasticSearchDomain.DomainEndpoint]]
+    OAUTH2_DOMAIN_ENDPOINT:
+      !Join ['', ['https://', !Ref UserPoolDomain, !Sub '.auth.${AWS::Region}.amazoncognito.com/oauth2']]
+  apiKeys:
+    - name: 'developer-key-${self:custom.stage}' # Full name must be known at package-time
```
rsmayda, line 40: Do you want to use !Ref Stage here as well?

serverless.yaml
```diff
@@ -1,362 +1,418 @@
...
+        Type: String
+        Default: ""
+        Description: The ARN of the IAM user to use for KMS key management. Ignored if ShouldCreateManagementUser is true.
+      Stage:
+        Type: String
+        Default: ${self:custom.stage}
+        Description: "The deployment stage (e.g. dev, qa, prod). Default: dev"
+      KibanaUserEmail:
+        Type: String
+        Default: abc@xyz.com
```
rsmayda: Should there be a default?

serverless.yaml
```diff
@@ -1,362 +1,418 @@
...
+resources:
+  - Parameters:
+      ShouldCreateManagementUser:
+        Type: String
+        Default: true
+        Description: "Set to 'true' to create management IAM user, 'false' if providing user ARN. Default: true"
+        AllowedValues: [ true, false ]
+      UserPassword:
+        Type: String
+        Default: DevPassword123@
```
rsmayda: We should not have a default here

rsmayda (commented):
To fix unit tests please commit an empty commit: https://coderwall.com/p/vkdekq/git-commit-allow-empty

cloudformation/alarms.yaml
swoldemi: [One of the dimensions in this alarm does a `!Ref DdbToEsLambdaFunction`](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/88c9574338f48dc0a7241de5e5f6856061335cac/cloudformation/alarms.yaml#L19). The dependency is made that way. Was just one of the linter warnings that cfn-lint mentioned so I removed it

cloudformation/elasticsearch.yaml
swoldemi: Agreed, I was just seeing how adding a name to the ES cluster would affect how the Kibana User Pool would work. Will remove

scripts/s3-dist.sh
swoldemi: $0 just references the first segment of the command (the script path and name), in case it gets renamed from s3-dist.sh to something else. 
If the script is ran as `$ ./scripts/asdasdasd.sh my-bucket`, then this prints and exits since both args are required for the script to run: 

```
Please provide both the prefix and region for the bucket
For example: ./scripts/asdasdasd.sh solutions us-east-1
```

scripts/s3-dist.sh
swoldemi: [Line 33](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/88c9574338f48dc0a7241de5e5f6856061335cac/scripts/s3-dist.sh#L33) adds provider.deploymentBucket to line 20. This prevents `sls package` from creating a default bucket, and using the parameters provided to the script instead. When the script exits, it removes it here (trapping the EXIT signal on [line 34](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/88c9574338f48dc0a7241de5e5f6856061335cac/scripts/s3-dist.sh#L34)), incase developers are already using the default bucket. 

It's not very portable since any changes that may move line 20 outside of the `provider` segment of the serverless.yaml will break this. I couldn't think of a better way for the Solutions team to provide their own bucket. Open to other suggestions.

serverless.yaml
swoldemi: `sls package` creates a Usage Plan for the API Key. Using !Ref Stage, doesn't work because Stage is not known to `sls package`, only variable (not Parameter) values are replaced when `sls package` is run. Doing !Ref Stage prevents `sls package` from completing with this error:

```
...
Serverless: Packaging service...
Serverless: Excluding development dependencies...

  Serverless Error ---------------------------------------

  API key "name" has no usage plan defined

  Get Support --------------------------------------------
...
```

serverless.yaml
swoldemi: Good point, will remove it

cloudformation/elasticsearch.yaml
```diff
@@ -127,4 +157,11 @@ Resources:
               Action: 'es:*'
               Resource:
                 - !Sub 'arn:aws:es:${AWS::Region}:${AWS::AccountId}:domain/*'
+            - Action: es:*
+              Principal:
+                AWS:
+                  - !Join ["/", [ !Sub 'arn:aws:sts::${AWS::AccountId}:assumed-role', !Ref KibanaCognitoRole, CognitoIdentityCredentials]]
```
ferdingler: When defining ARNs, its better to use `${AWS::Partition}` instead of hardcoding `aws` because special regions like GovCloud or China have different partition names (i.e. aws-gov, aws-cn, etc). 

```
arn:${AWS::Partition}:sts::${AWS::AccountId}:assumed-role
```

scripts/s3-dist.sh
rsmayda: Check this out I think this is what we want? https://medium.com/serverlessguru/how-to-reuse-an-aws-s3-bucket-for-multiple-serverless-framework-deployments-d1673d3d8259

serverless.yaml
```diff
@@ -1,362 +1,417 @@
...
+      CognitoOAuthRedirectURL:
+        Type: String
+        Default: ${self:custom.oauthRedirect}
+        Description: "Cognito OAuth redirect URL used for User Pool. Default:  ${self:custom.oauthRedirect}"
+  - Conditions:
+      isDev: !Equals [!Ref Stage, 'dev']
+      isNotDev: !Not [Condition: isDev]
+      CreateManagementUser: !Equals [!Ref ShouldCreateManagementUser, true]
+  - Resources:
+      ManagementUser:
```
rsmayda: Just to understand a bit more what is the goal of having this user/why did you add it? is it a best practice?

swoldemi: The installation/setup guide says to provide a devAwsUserAccountArn in the serverless_config.json file for the template to read. Since CloudFormation won't know of the custom variables provided to `sls package` and can't read Parameter values from JSON files, I added a Parameter for customers to provide the ARN. If they would prefer that the template create the user with the [recommended permissions](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/scripts/iam_policy.json), then they can set this flag to true. 

If it's better to just have customers only provide the IAM user ARN parameter, we can remove the user and do that too.

scripts/s3-dist.sh
swoldemi: This blog is what lead me to this solution actually, using provider.deploymentBucket to use an existing bucket. This way the Solutions team can provide their own existing buckets (`solutions-<region-name>`) at build time (when they run the script) so customers can deploy the template into their own accounts. Without this somewhat hacky automation, the Solutions team would have to provide the deploymentBucket parameter to the serverless.yaml template themselves. Or was there something else in the blog that I missed?

rsmayda: > Without this somewhat hacky automation, the Solutions team would have to provide the deploymentBucket parameter to the serverless.yaml template themselves.

What is wrong with that? We could create a new serverless CLI variable 'deployBucket' and pass that in?

`sls package --region "$BUCKET_REGION" --deployBucket '"$BUCKET_PREFIX"'-'"$BUCKET_REGION"'`

https://www.serverless.com/framework/docs/providers/aws/guide/variables/#referencing-cli-options

Basically can we remove `sed` cmds and just use sls?

swoldemi: Passing --deployBucket means provider.deploymentBucket already exists in the template, right? This would work, as long as developers that will be running sls commands are okay with always providing a deploymentBucket. Or is there an sls native way to ignore provider.deploymentBucket if --deployBucket isn't provided? 

The issue the `sed` commands try to solve is more allowing the Solutions team to be able to package and distribute the template to their own existing buckets while also giving the FHIR Works development team the flexibility of continuing to use the default bucket provided by the sls framework.

rsmayda: > deployBucket means provider.deploymentBucket already exists in the template, right?

Actually it is worse it means that the bucket exists in the account.

hmmm ok so it sounds like this is the best approach. I think at the very least we need to remove the hard coded 19/20 line number. Could you look into a way to get the line number `providers:` is on?

Maybe this is a good lead: https://stackoverflow.com/a/18870500/9041033


Similar problem: https://forum.serverless.com/t/how-to-create-deploymentbucket-when-manually-specified-if-it-doesnt-exist/2439

serverless.yaml
rsmayda: My vote is to remove this and have the customer's be responsible for the management aspect. @nguyen102 @carvantes  any opinion here?

cloudformation/backup.yaml
```diff
@@ -4,7 +4,7 @@
 #
 
 # NOTE: this is an extra Cloudformation stack; not linked to FHIR Server stack
-Description: 'Backup Plan template to back up all resources tagged with backup=daily daily at 5am UTC.'
+Description: '(SO0128) - Solution - Backup Plan Template - This template creates resources necessary to back up all resources tagged with backup=daily daily at 5am UTC.'
```
nguyen102, line 7: Out of curiosity, what `(SO0128)` does stand for?

serverless.yaml
nguyen102: Either way the customer needs to provide the template with an IAM User Arn, right? If we can automate that by having this `ManagementUser`, it seems that this would make the installation process simpler for the customer.

rsmayda: I think this management user is purely for managing the created AWS resources from this template, not for the deployment. If it was required for deployment it would need to be created prior to the cfn template

serverless.yaml
```diff
@@ -1,362 +1,417 @@
...
+        Description: "Set to 'true' to create management IAM user, 'false' if providing user ARN. Default: true"
+        AllowedValues: [ true, false ]
+      UserPassword:
+        Type: String
+        Description: Password for development IAM user. Ignored if ShouldCreateManagementUser is false.
+        NoEcho: true
+      ExistingManagementUserARN:
+        Type: String
+        Default: ""
+        Description: The ARN of the IAM user to use for KMS key management. Ignored if ShouldCreateManagementUser is true.
```
nguyen102: When running `sls offline` to allow local debugging of the FHIR server, serverless uses the credentials of this user. ES is set to allow this user to make requests to ES.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/5780f97afe427b26405d82062fbc86b7f61d43a8/cloudformation/elasticsearch.yaml#L149

serverless.yaml
nguyen102: Looks like this IAM user is used to manage the KMS keys and to allow customers running `sls offline` to [access ES](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/5780f97afe427b26405d82062fbc86b7f61d43a8/cloudformation/elasticsearch.yaml#L149). 

I guess it would be simpler if we ask the user to provide the IAM user they're using to deploy the FHIR server, instead of creating a new IAM user for them.

carvantes: Do not create the `ManagementUser`.

Also, the KMS/ES resource based policies don't really need to mention `devAwsUserAccountArn`/`ManagementUser`. They are sort of redundant. Not sure about exactly why such statement was added in the first place.

If you are deploying/offline-testing FWoA with a role user/role that has and identity police that allows `kms:*`, `es:*` (as suggested on the iam_policy.json file) then removing the above KMS/ES statements has no effect.

scripts/s3-dist.sh
```diff
@@ -25,17 +25,20 @@ BUCKET_PREFIX=$1
 BUCKET_REGION=$2
 
 function cleanup(){
```
rsmayda, line 27: weird not used to shell scripts you don't pass in a param, but instead use $1?

swoldemi: $1 within the scope of `cleanup` is different from the scope of the rest of the script because a param is passed to the `trap` call [here](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/5635531ad79cdf0fbfcf50546b71999f392b8f58/scripts/s3-dist.sh#L41)

rsmayda: Interesting TIL :) more reading I checked into: https://www.shellscript.sh/functions.html

serverless.yaml
```diff
@@ -10,18 +10,20 @@ custom:
   resourceTableName: 'resource-db-${self:custom.stage}'
   stage: ${opt:stage, self:provider.stage}
   region: ${opt:region, self:provider.region}
+  kibanaUserEmail: ${opt:kibanaUserEmail, 'abc@xyz.com'}
```
rsmayda: I don't think we want a default Kibana user email address

serverless.yaml
```diff
@@ -10,18 +10,20 @@ custom:
...
   oauthCallback: ${opt:oauthCallback, 'http://localhost'}
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
 
 provider:
   name: aws
+  deploymentBucket: swoldemi-us-east-1
```
rsmayda: I think this is accidental code

serverless.yaml
```diff
@@ -140,6 +142,7 @@ resources:
         Description: "The deployment stage (e.g. dev, qa, prod). Default: dev"
       KibanaUserEmail:
         Type: String
+        Default: ${self:custom.kibanaUserEmail}
```
rsmayda: I think this is okay but not setting a default on sls side. Also I think it is fine if we don't initialize the Kibana user automatically -- the kibana experience is very niche and only for developers

serverless.yaml
swoldemi: Good catch, will remove

serverless.yaml
swoldemi: Will replace it with an empty string

serverless.yaml
swoldemi: Should we just remove all the Kibana User Pool user creation stuff then?

rsmayda: We slacked, but yes we should!

cloudformation/elasticsearch.yaml
```diff
@@ -1,130 +1,135 @@
...
+                AWS:
+                  - !GetAtt AdminKibanaAccessRole.Arn
+              Action: 'es:*'
+              Resource:
+                - !Sub 'arn:${AWS::Partition}:es:${AWS::Region}:${AWS::AccountId}:domain/*'
+            - Action: es:*
+              Principal:
+                AWS:
+                  - !Join ["/", [ !Sub 'arn:${AWS::Partition}:sts::${AWS::AccountId}:assumed-role', !Ref KibanaCognitoRole, CognitoIdentityCredentials]]
+              Effect: Allow
```
rsmayda: Nitpick: the order of the statements are inconsistent

can it be 
* Effect
* Principal
* Action
* Resource

cloudformation/kms.yaml
```diff
@@ -21,22 +21,13 @@ Resources:
           - Sid: 'Enable IAM Root Permissions'
             Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
+              AWS: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:root'
             Action: 'kms:*'
             Resource: '*'
-          - Fn::If:
-              - isDev
```
rsmayda, line 28: @nguyen102 If this is removed will we be able to use KMS when we do `sls offline`?

cloudformation/kms.yaml
```diff
@@ -50,22 +41,13 @@ Resources:
...
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
+              AWS: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:root'
             Action: 'kms:*'
             Resource: '*'
-          - Fn::If:
-              - isDev
-              - Sid: 'Enable IAM User Permissions'
-                Effect: Allow
-                Principal:
-                  AWS: ${self:custom.config.devAwsUserAccountArn}
```
rsmayda, line 61: @swoldemi are you removing this stuff since it can't be determined by cloudformation?

rsmayda (commented):
More small nitpicks -- almost there

cloudformation/kms.yaml
nguyen102: The user role that is running `sls offline` should still be able to access the S3 bucket if the user role has KMS access permission. So if you're using an admin role, then `sls offline` should still work.

It looks like this KMS key already provide access to the root account.
More info here.
https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html

cloudformation/kms.yaml
swoldemi: I removed it on @carvantes's recommendation in [an earlier comment](https://github.com/awslabs/fhir-works-on-aws-deployment/pull/121#discussion_r504298286)

rsmayda: Yep thanks for the clarification

rsmayda (approved):
Looks good! Thanks for your hard work

PR #122, opened 2020-10-15
feat: add lambda fns to start glue jobs and check its status

carvantes:
Description of changes:

Add CFN resources to support the startExportJob and getJobStatus lambda functions.

ExportGlueJob and GlueJobRole may need some changes once we have the actual Glue script. For now i tested the permissions/config with the most basic script that's generated from the Glue console(pretty much just dump DDB to S3 as is)

Checklist:

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/27

### DIFF
--- a/cloudformation/bulkExport.yaml
+++ b/cloudformation/bulkExport.yaml
@@ -118,7 +118,7 @@ Resources:
         Rules:
           - Id: ExpirationRule
             Status: Enabled
-            ExpirationInDays: '7'
+            ExpirationInDays: '3'
 
   GlueJobRole:
     Type: AWS::IAM::Role
--- a/cloudformation/bulkExport.yaml
+++ b/cloudformation/bulkExport.yaml
@@ -77,3 +77,96 @@ Resources:
                 Action:
                   - glue:StartCrawler
                 Resource: !Join ['', [!Sub 'arn:aws:glue:${AWS::Region}:${AWS::AccountId}:crawler/', !Ref Crawler]]
+
+  GlueJobRelatedLambdaRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service: 'lambda.amazonaws.com'
+            Action: 'sts:AssumeRole'
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+        - arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess
+      Policies:
+        - PolicyName: glueAccess
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - glue:StartJobRun
+                  - glue:GetJobRun
+                Resource: '*'
+              - Effect: Allow
+                Action:
+                  - dynamodb:Query
+                  - dynamodb:GetItem
+                Resource:
+                  - !GetAtt ExportRequestDynamoDBTable.Arn
+
+  GlueScriptsBucket:
+    Type: AWS::S3::Bucket
+
+  BulkExportResultsBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      LifecycleConfiguration:
+        Rules:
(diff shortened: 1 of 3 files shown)


cloudformation/bulkExport.yaml
```diff
@@ -77,3 +77,96 @@ Resources:
...
+    Type: AWS::S3::Bucket
+
+  BulkExportResultsBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      LifecycleConfiguration:
+        Rules:
+          - Id: ExpirationRule
+            Status: Enabled
+            ExpirationInDays: '7'
```
nguyen102: I think the items in the bucket only need to stay in the bucket for as long as the presigned Url for the exported filles are valid. It's unlikely the presigned url is going to last 7 days. 

Maybe we can set a custom variable value in serverless for how long the presigned S3 url is going to last. Provide that as an environment variable so we can generate a presigned S3 url for that given length of time. Then for the S3 Bucket policy have the expiration time equal to presigned S3 url expiration time + 15 minutes. We can exchange 15 minutes for an hour to be safe.

rsmayda: Could they request another presigned URL for the same export file? Or are we only support 1 and done?

cloudformation/bulkExport.yaml
```diff
@@ -77,3 +77,96 @@ Resources:
...
+                  - s3:GetObject
+                Resource: !Join ['', ['arn:aws:s3:::', !Ref GlueScriptsBucket,'/*']]
+
+  ExportGlueJob:
+    Type: AWS::Glue::Job # TODO: update this configuration once we have an actual ETL script
+    Properties:
+      Role: !GetAtt GlueJobRole.Arn
+      Command:
+        ScriptLocation: !Join ['', ['s3://', !Ref GlueScriptsBucket, '/our-glue-export-script.py']]
+        Name: glueetl
```
rsmayda, line 171: Do we require setting the name? looking at the configuration https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-glue-job.html#cfn-glue-job-name
Update requires replacement. 

I generally stay away from using AWS names and just use ARNs

serverless.yaml
```diff
@@ -163,6 +163,26 @@ functions:
...
 
+  startExportJob:
+    timeout: 30
+    memorySize: 192
+    runtime: nodejs12.x
+    description: 'Start the Glue job for bulk export'
+    role: GlueJobRelatedLambdaRole
+    handler: bulkExport/index.startExportJobHandler
+    environment:
+      GLUE_JOB_NAME: !Ref ExportGlueJob
```
rsmayda, line 174: safer to use arn instead?

cloudformation/bulkExport.yaml
nguyen102: The original tech design is we create the presigned s3 url once the export is complete. We store that in DDB and retrieve that when a user asks for the files. In this approach a new url is not generated.

Nestor and I discussed offline about allowing the url to be regenerated though. I believe Nestor is going to send an email to the group to discuss the new approach. And we can confirm with Angus/Steven that the new approach would still be ok for our customers.

cloudformation/bulkExport.yaml
carvantes: This is not the job name, It is the Command name which is actually more like an enum.
>The name of the job command. For an Apache Spark ETL job, this must be glueetl. For a Python shell job, it must be pythonshell.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-glue-job-jobcommand.html

serverless.yaml
carvantes: why safer?

Job name is sufficient. It is assumed that the job is on the same account and region. Similar to how we only pass ddb table names as env variables.

cloudformation/bulkExport.yaml
carvantes: Changed to 3 days

carvantes: We are still having conversations about the exact details around presigned s3 urls. 
I may make an adjustment to this bucket lifecycle policy later on when working on the presigned urls story. It's probably fine to not block this PR on this detail.

nguyen102 (approved):

PR #125, opened 2020-10-21
feat: Add OAuth2 support for SMART on FHIR

rsmayda:
Description of changes:
First part of OAuth2 support for SMART on FHIR. We are using /userinfo endpoint to get fhirUser claim. Three parts are verified:
1. Validity of `aud` and `iss` jwt information
1. Validity that the incoming scopes allow requested operations
1. Validity of JWT (via the /userinfo endpoint)

AuthZ of scopes is based on the rules found in authZConfig.ts file. Effectively certain scopes allow for certain operations. for example: `patient/Patient.write` does allow creates while `user/Patient.write` does not. These are customer specific decisions and our defaults

Checklist:


serverless.yaml
```diff
@@ -10,8 +10,7 @@ custom:
   resourceTableName: 'resource-db-${self:custom.stage}'
   stage: ${opt:stage, self:provider.stage}
   region: ${opt:region, self:provider.region}
-  oauthCallback: ${opt:oauthCallback, 'http://localhost'}
-  oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
+  oauthDomain: ${opt:oauthCallback, 'https://dev-6460611.okta.com/oauth2/default'}
```
mgresko: should there be a specific okta reference here?

mgresko:
@rsmayda i thought smart-on-fhir was going to be a separate package?  If we think about specific implementations and deviating from the aws-deployment it will be difficult to pull in these changes.

serverless.yaml
rsmayda: Good callout this should be a less precise default

rsmayda:
> @rsmayda i thought smart-on-fhir was going to be a separate package? If we think about specific implementations and deviating from the aws-deployment it will be difficult to pull in these changes.

It will :) for now we are adopting this branching approach until we are able to create a new private package

cloudformation/elasticsearch.yaml
```diff
@@ -88,21 +88,21 @@ Resources:
       EBSOptions: # Assuming ~100GB storage requirement for PROD; min storage requirement is ~290GB https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/sizing-domains.html
         EBSEnabled: true
         VolumeType: gp2
-        VolumeSize: !If [isDev, 10, 73]
+        VolumeSize: !If [isNotProd, 10, 73]
```
nguyen102: Nice!

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,130 @@ import {
...
+    SystemOperation,
+    TypeOperation,
 } from 'fhir-works-on-aws-interface';
-import { SMARTConfig } from './smartConfig';
+import axios from 'axios';
+import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
 export class SMARTHandler implements Authorization {
+    static readonly SCOPE_REGEX = /(patient|user|system|launch)\/(\w+|\*)\.?(\w+|\*)?|(^launch$)/;
```
nguyen102: In SMART are these the only four possible scopes? Would it be possible for `scopeRule` to have more than these 4 types, and therefore cause this regex to become invalid?

src/authz-smart/smartConfig.ts
```diff
@@ -2,6 +2,35 @@
  *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *  SPDX-License-Identifier: Apache-2.0
  */
+import { TypeOperation, SystemOperation } from 'fhir-works-on-aws-interface';
+
```
nguyen102, line 6: Normally we have our Config interfaces in `fhir-works-on-aws-interface`. For example `FHIRConfig` and `Auth`. What are your thoughts on putting it in the `interface` package vs putting it here?

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,130 @@ import {
...
+        }
+        // verify scope
+        let scopes: string[] = [];
+        if (this.config.scopeValueType === 'space' && typeof decoded[this.config.scopeKey] === 'string') {
+            scopes = decoded[this.config.scopeKey].split(' ');
+        } else if (this.config.scopeValueType === 'array' && Array.isArray(decoded[this.config.scopeKey])) {
+            scopes = decoded[this.config.scopeKey];
+        }
+        if (!this.isScopeSufficient(scopes, request.operation, request.resourceType)) {
+            console.error('scopes are insuffiecient');
```
nguyen102: Is there a reason these errors are broken into 4 statements? Can they be combined into one statement?

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,130 @@ import {
...
+            console.error(`resourceType: ${request.resourceType}`);
+            return false;
+        }
+
+        // Verify token
+        if (this.config.authZUserInfoUrl) {
+            console.log('Posting to AuthZ for more customer information');
+            let result;
+            try {
+                result = await axios.post(
```
nguyen102: nitpick: `result` => `response`

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,130 @@ import {
...
+                    this.config.authZUserInfoUrl,
+                    {},
+                    { headers: { Authorization: `Bearer ${request.accessToken}` } },
+                );
+            } catch (e) {
+                console.error('Post to authZUserInfoUrl failed');
```
nguyen102: Can this be combined to one line?
`console.error('Post to authZUserInfoUrl failed', e);`

serverless.yaml
```diff
@@ -145,6 +127,7 @@ functions:
 
 resources:
   - Conditions:
+      isNotProd: !Not !Equals ['${self:custom.stage}', 'prod']
       isDev: !Equals ['${self:custom.stage}', 'dev']
```
carvantes: Why do we have both `isNotProd` and `isDev`?

rsmayda: My distinction is we only want the expensive elastic search if the users use stage 'prod'. perhaps better variable name would help?

src/authz-smart/smartConfig.ts
rsmayda: Yeah so this will be much like how we set up RBAC (https://github.com/awslabs/fhir-works-on-aws-authz-rbac/blob/mainline/src/RBACConfig.ts)

It is a configuration that doesn't apply widely just to the SMART implementation

src/authz-smart/smartHandler.ts
rsmayda: the first part of the scope currently can only have 4 (this may change as the spec changes). I think this regex is complete - but would love to know if you see an errror

Clinical scopes: http://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context/#clinical-scope-syntax
Launch scopes: http://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context/#requesting-context-with-scopes

src/authz-smart/smartHandler.ts
rsmayda: Nope no reason -  can do!

nguyen102 (dismissed):

serverless.yaml
carvantes: What was wrong with using `isDev` to keep ES small?

I think we should be careful about supporting more "special stage names" like 'dev', 'prod'. The fewer the better.

Also, maybe we should not make this change as part of this PR anyways. It's not related to SMART and Idk if it will reach mainline given the branching strategy for SMART.

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,130 @@ import {
...
+                console.error(e);
+            }
+            if (!result || !result.data[this.config.expectedFhirUserClaimKey]) {
+                console.error(`result from AuthZ did not have ${this.config.expectedFhirUserClaimKey} claim`);
+                return false;
+            }
+        }
         return true;
     }
```
carvantes: Are we ok with making a call to /userInfo on every request? This is expensive

Isn't it possible to have the FhirUser claim in the access token to begin with and validate the token signature without a /userInfo call?

Also, this code has a weird order in that it analyzes claims/scopes of a possibly invalid token and then the token is validated w /userInfo at the end. Validating the token should be the very first thing.

carvantes (changes requested):

src/authz-smart/smartHandler.ts
carvantes: All the JWT in the unit tests do have the `fhirUser` claim in them.

src/authz-smart/smartHandler.ts
carvantes: What is the system scope?

Why is`'read' | 'write'` not explicitly on the regex? `patient/Observation.doSomething` shouldn't match

Use anchors `^$`, `SomePreffixpatient/Patient.readSomeSuffix` should not match 

It may be better to use separate regex for clinical scopes and launch scopes or separate them in a more explict way (like `<clinical regex>|<launch regex>`). Right now this matches invalid weird combinations like `launch/*.read`

It may be worth having unit tests specifically for the `isScopeSufficient` function

serverless.yaml
rsmayda: Good point encapsulating PRs is a best practice - reverting

src/authz-smart/smartHandler.ts
rsmayda: Their are two ways to validate JWTs 
1. Decoding the JWT and verifying signature
    * We will need to call the AuthZ server for the public key - we could then cache that public key (outside of lambda)
    * This leaves the vulnerability of AuthZ revoking the accessToken and the FHIR API still accepting it as valid
1. Using Token introspection or userInfo endpoint
    * Downside here is network call for each AuthZ
    * We could move this to a custom authorizer (where we could cache) but same issue as above there are race conditions that would allow revoked accessTokens to still gain access
    * As a note the userInfo endpoint validates AND return user's claims

Yeah you are right you can add the fhirUser to the JWT access_token the trouble is you still need to verify it

As for code ordering, my thought is do the local checks first then do the expensive call -- I think ideally we use promises: kick off the call do local checks then check the response

https://developer.okta.com/docs/guides/validate-access-tokens/nodejs/overview/

src/authz-smart/smartHandler.ts
rsmayda: System scope -> http://build.fhir.org/ig/HL7/bulk-data-export/authorization/index.html#scopes

+1 on regex fixes

Adding unit tests for isScopeSufficient means it will need to be public (not the end of the world but a bit odd)

src/authz-smart/smartHandler.ts
carvantes: Why are we using `/userinfo` instead of `/introspect`?
`/introspect` is meant to validate tokens. 
`/userinfo` Is meant to provide user info and validates tokens as a side effect. Additional permissions scopes are required to call it and you are getting user info that the fhir server doesn't need (i.e. `family_name`)

In general designing for token revocation is a risk/performance balance. longer cache time and longer lived tokens give better performance but increase the risk of letting a revoked token go through.
Is the goal here to have zero tolerance for revoked tokens(and consequently the worst possible performance)?

Is this a generic implementation? Or is this Okta specific? Not all idps can revoke tokens to begin with.

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,135 @@ import {
...
+    SystemOperation,
+    TypeOperation,
 } from 'fhir-works-on-aws-interface';
-import { SMARTConfig } from './smartConfig';
+import axios from 'axios';
+import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
 export class SMARTHandler implements Authorization {
+    static readonly CLINICAL_SCOPE_REGEX = /^(patient|user|system)\/(\w+|\*)\.(read|write|\*)$/;
```
carvantes: `w+` may be too open. I believe ResourceTypes are letters only
`patient/Observation_77.read` shouldn't match

src/authz-smart/smartHandler.ts
rsmayda: > Why are we using /userinfo instead of /introspect?

Some customer may not want to add `fhirUser` to the `access_token`, as it does leak **some** customer information. The alternative to adding it to the `access_token` is to use the `/userinfo` endpoint to ensure we get the `fhirUser` claim. Without this claim we do not know who they are.

> Is the goal here to have zero tolerance for revoked tokens(and consequently the worst possible performance)?

The goal with this PR is to solve the problem functionally first then optimize if there is customer interest/demand. An easy path of optimization is adding this logic into a Custom Authorizer and let API GW cache this information per JWT token.

> Is this a generic implementation? Or is this Okta specific? Not all idps can revoke tokens to begin with.

This should be a generic OAuth2/OIDC implementation. If you see Okta specific concerns please let me know!

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,136 @@ import {
...
+        }
+        if (!this.isScopeSufficient(scopes, request.operation, request.resourceType)) {
+            console.error(
+                `User supplied scopes are insuffiecient\nscopes: ${scopes}\noperation: ${request.operation}\nresourceType: ${request.resourceType}`,
+            );
+            return false;
+        }
+
+        // Verify token
+        if (authZPromise) {
```
carvantes: why `if`? authzPromise is always truthy here

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,136 @@ import {
...
     }
+
+    private isScopeSufficient(
+        scopes: string[],
+        operation: TypeOperation | SystemOperation,
+        resourceType?: string,
+    ): boolean {
+        const { scopeRule } = this.config;
+        let isAuthorized = false;
+        for (let i = 0; i < scopes.length && !isAuthorized; i += 1) {
```
carvantes: I don't like the mutation of `isAuthorized` and  having it as a condition in the for loop (`!isAuthorized`)

what are your thoughts and just returning true the moment we find a match with valid operations, otherwise just return false at the very end of the function?

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,136 @@ import {
...
+            const scope = scopes[i];
+            let results = scope.match(SMARTHandler.LAUNCH_SCOPE_REGEX);
+            if (!results) {
+                results = scope.match(SMARTHandler.CLINICAL_SCOPE_REGEX);
+            }
+            if (results !== null && results.length > 3) {
```
carvantes: why check `results.length`? we know that there's a fixed number of capture groups on the regex

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,136 @@ import {
...
+                const scopeType: string = results[1];
+                let validOperations: (TypeOperation | SystemOperation)[] = [];
+                if (scopeType === 'launch') {
+                    const launchType: string = results[3];
```
carvantes: Not mandatory, but you could use named capture groups to make the code clearer. [example](https://github.com/awslabs/fhir-works-on-aws-search-es/blob/29a11ef764819957acfda7e901a429852ed4330f/src/searchInclusions.ts#L33)

Or at the very least use non capturing groups for the parentheses that we don't want to capture(I was initially confused about why it was 3 instead of 2 here)

src/authz-smart/smartHandler.ts
```diff
@@ -8,31 +8,136 @@ import {
...
+            { headers: { Authorization: `Bearer ${request.accessToken}` } },
+        );
+
+        const decoded = decode(request.accessToken, { json: true }) || {};
+        const { aud, iss } = decoded;
+        // verify aud & iss
+        if (this.config.expectedAudValue !== aud || this.config.expectedIssValue !== iss) {
+            console.error('aud or iss is not matching');
+            return false;
+        }
```
carvantes: I *think* that the authZPromise promise can crash the server since it ends up as an unhandled rejection if this function returns false early AND the call to the auth server fails.

Regarding the code order, after the discussion I think that it was better as it was originally written. I guess it's fine to check some claims like aud first so that some bad tokens are rejected early without needing the call to `/userinfo`.

Maybe just add a comment above the call to jwt.decode mentioning that the token will be validated by the /userInfo call. It's just that using `jwt.decode` instead of `jwt.verify` is usually a massive red flag that can confuse people reading this code, but here the overall implementation is correct.

carvantes (changes requested):

src/authz-smart/smartHandler.ts
carvantes: discussed offline a bit more. I'm ok with the `/userinfo` call here

src/authz-smart/smartHandler.ts
rsmayda: Yep that works for me

src/authz-smart/smartHandler.ts
rsmayda: +1 left over from previous iterations

src/authz-smart/smartHandler.ts
rsmayda: I like this - good tips on regexes thank you

src/authz-smart/smartHandler.ts
rsmayda: interesting when running the unit tests I didn't see those sort of failures but good callout it may be one

src/config.ts
```diff
@@ -36,23 +45,16 @@ export const fhirConfig: FhirConfig = {
...
-                    : process.env.OAUTH2_DOMAIN_ENDPOINT,
+            service: 'SMART-on-FHIR',
+            oauthUrl,
         },
     },
     server: {
         // When running serverless offline, env vars are expressed as '[object Object]'
         // https://github.com/serverless/serverless/issues/7087
         // As of May 14, 2020, this bug has not been fixed and merged in
         // https://github.com/serverless/serverless/pull/7147
```
carvantes: Move this comment up. It no longer makes sense here

carvantes (dismissed):

carvantes (approved):

PR #127, opened 2020-10-26
feat: add s3 presigned urls to bulk export status response

carvantes:
Description of changes:
Add resources and env variables for signing bulk export results urls.

Checklist:

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/21

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -40,6 +40,10 @@ provider:
     ELASTICSEARCH_DOMAIN_ENDPOINT: !Join ['', ['https://', !GetAtt ElasticSearchDomain.DomainEndpoint]]
     OAUTH2_DOMAIN_ENDPOINT:
       !Join ['', ['https://', !Ref UserPoolDomain, '.auth.${self:custom.region}.amazoncognito.com/oauth2']]
+    EXPORT_RESULTS_BUCKET:
+      Ref: BulkExportResultsBucket
+    EXPORT_RESULTS_SIGNER_ROLE_ARN:
+      !GetAtt ExportResultsSignerRole.Arn
   apiKeys:
     - name: 'developer-key-${self:custom.stage}'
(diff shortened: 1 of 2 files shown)


serverless.yaml
```diff
@@ -96,6 +100,18 @@ provider:
       Effect: Allow
       Resource:
         - '*'
+    - Action:
+        - 's3:ListBucket'
```
nguyen102, line 104: Is this to allow creating presigned S3 urls? Is the ListBucket policy needed for creating presigned S3 urls?

carvantes: No, there's a role `ExportResultsSignerRole` that's used specifically for signing urls. We want to assume a different role so that we have control over the role session duration(otherwise the urls may stop working before the 30 min expiration time if the lambda role session expires before that).

The main lambda function needs `s3:ListBucket` to find all the result files for a given JobId

nguyen102: Sounds good!

nguyen102 (approved):

serverless.yaml
```diff
@@ -96,6 +100,18 @@ provider:
       Effect: Allow
       Resource:
         - '*'
+    - Action:
+        - 's3:ListBucket'
+        - 's3:GetObject'
+      Effect: Allow
+      Resource:
+        - !GetAtt BulkExportResultsBucket.Arn
+        - !Join ['', [!GetAtt BulkExportResultsBucket.Arn, '/*']]
```
rsmayda, line 109: Just for my learning why do we need both here?

carvantes: First one is a Bucket arn. Second one is an Object arn (with a wildcard).
`s3:ListBucket` operates on Buckets.  `s3:GetObject` operates on Objects. Both resources are needed to make the actions valid

rsmayda (approved):

PR #128, opened 2020-10-26
Export job config

nguyen102:
Description of changes:
* Added Glue script for exporting DDB to S3 as a reference file. (TODO: Figure out how to upload Glue script to S3 as part of CloudFormation)
* Glue uses Apache Spark, which export files in the format `<jobId>/resourceType=<resourceType>/run- run-1603724177164-part-r-<5 digit number from 0 to n>`. Based on this [recommendation](https://ujjwalbhardwaj.me/post/rename-aws-glue-job-output-file), let's rename the files. To do this we need IAM permission to GET and DELETE objects in `BulkExportResultsBucket`. 

Checklist:


scripts/export-script.py
```diff
@@ -0,0 +1,94 @@
...
+    # Rename exported files into ndjson file. One file for each resourceType
+    client = boto3.client('s3')
+
+    response = client.list_objects(
+        Bucket=bucket_name,
+        Prefix=job_id,
+    )
+
+    regex_pattern = '\/resourceType=(\w+)\/'
+    for item in response["Contents"]:
```
nguyen102: source_file_path looks like this `<jobId>/resourceType=<resourceType>/run- run-1603724177164-part-r-<5 digit number from 0 to n>`

We're copying it to a new file name, and then deleting the original source file.

scripts/export-script.py
```diff
@@ -0,0 +1,94 @@
+import sys
```
carvantes: I think we should move this file to a different location. The `scripts` folder is for scripts that customers can run locally. 

Maybe create another scripts folder under /bulkExport

scripts/export-script.py
```diff
@@ -0,0 +1,94 @@
...
+transaction_time = args['transactionTime']
+since = args['since']
+outputFormat = args['outputFormat']
+glue_table_name = args['glueTableName']
+
+glue_database = args['glueDatabase']
+bucket_name = args['s3OutputBucket']
+
+# Read data from DDB
+original_data_source_dyn_frame = glueContext.create_dynamic_frame.from_catalog(database = glue_database, table_name = glue_table_name)
```
carvantes: Were you testing this script on the Glue console?
Can you add the annotations like `## @type: DataSource` so that it generates the execution plan diagram? 

https://docs.aws.amazon.com/glue/latest/dg/edit-script.html

carvantes (changes requested):

scripts/export-script.py
nguyen102: For most of the development of the script, I used `sagemaker` to develop and test the script. [Link here](https://aws.amazon.com/about-aws/whats-new/2018/10/aws-glue-now-supports-connecting-amazon-sagemaker-notebooks-to-development-endpoints/)

One the script is working in Sagemaker I ported it to the Glue console for the final test. I don't think the annotations are strictly necessary. It's more helpful if you're starting out from a Glue template and are making minor changes. The annotation is then helpful if you want to visually see your changes using Glue's diagram.

For more complex scripts, I think comments in the code is enough. For example, the [sample glue scripts](https://github.com/aws-samples/aws-glue-samples/tree/master/examples) released by AWS doesn't have the diagram annotations.

scripts/export-script.py
nguyen102: Sure thing. Moving the code to the `bulkExport` folder.

bulkExport/export-script.py
```diff
@@ -0,0 +1,94 @@
...
+import boto3
+import re
+from awsglue.transforms import *
+from awsglue.utils import getResolvedOptions
+from pyspark.context import SparkContext
+from awsglue.context import GlueContext
+from awsglue.job import Job
+from awsglue.dynamicframe import DynamicFrame
+from datetime import datetime
+
```
rsmayda: Add a comment explaining what this file is doing and how it is used with glue and what functionality it is providing

Think of adding the information a forker of this repo will need to determine to keep this file or not for their deployment

nguyen102: updated

nguyen102: updated

bulkExport/export-script.py
```diff
@@ -0,0 +1,105 @@
+"""
```
carvantes: Actually, can you move this file to a folder dedicated to glue scripts? Maybe `bulkExport/glueScripts/export-script.py`

The thing is that pretty much all upload-to-s3-on-deployment upload the contents of a **folder** to s3

carvantes (changes requested):

bulkExport/export-script.py
nguyen102: yep, doing that now

bulkExport/export-script.py
```diff
@@ -0,0 +1,105 @@
...
+        Prefix=job_id,
+    )
+
+    regex_pattern = '\/resourceType=(\w+)\/'
+    for item in response["Contents"]:
+        source_s3_file_path = item["Key"]
+        match = re.search(regex_pattern, source_s3_file_path)
+        new_s3_file_name = match.group(1) + "-1.ndjson"
+        new_s3_file_path = job_id + "/" + new_s3_file_name
+        client.copy_object(Bucket=bucket_name, CopySource=bucket_name + "/" + source_s3_file_path, Key=new_s3_file_path)
```
carvantes: This will fail for files larger than 5GB. Use `copy` instead.

Also, in general I feel like this copy to rename step may end up being a bottleneck since it is done outside of Apache Spark. I'm not familiar with Glue, so I don't have any alternative in mind. 

Can you add log statements around this rename step so that we know how long it takes? I think that'll come in handy when running the tests for larde databases. 

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.copy
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.copy_object

carvantes (changes requested):

bulkExport/export-script.py
nguyen102: updated

cloudformation/bulkExport.yaml
```diff
@@ -166,7 +168,18 @@ Resources:
     Type: AWS::Glue::Job # TODO: update this configuration once we have an actual ETL script
     Properties:
       Role: !GetAtt GlueJobRole.Arn
+      GlueVersion: '2.0'
+      WorkerType: 'G.2X'
```
nguyen102, line 200: In testing for exporting 1 GB of data from DDB, I found that this `G.2X` workers resulted in the fastest export time for the number of DPU used. Our export task is memory intensive and G2 workers have the most memory.

bulkExport/glueScripts/export-script.py
```diff
@@ -0,0 +1,116 @@
...
+
+if filtered_dates_resource_dyn_frame.count() > 0:
+    print('Dropping fields not needed')
+    # Drop fields that are not needed
+    data_source_cleaned_dyn_frame = DropFields.apply(frame = filtered_dates_resource_dyn_frame, paths = ['documentStatus', 'lockEndTs', 'vid'])
+
+    print('Combining data into smaller partitions')
+    # Combine data into at most 10 partitions
+    data_frame = data_source_cleaned_dyn_frame.toDF()
+    data_frame = data_frame.coalesce(10)
```
nguyen102: Unfortunately, combining all of the data into 1 file is very slow and cause the Glue job to fail with an out of memory exception. Splitting the data into at most 10 files per `resourceType` allow the Glue job to export 1 GB of data in 10 minutes.

carvantes: why 10? can it be a static number?
I feel that it's going to fail with a similar out of memory error for for a DB that is 10x the size. But idk for sure

nguyen102: I did some further testing, and with G2x worker type, Glue can partition 1GB of data into 1 file per resource type. To keep things simple, I'm planning on setting a hard limit. When we have less than 100 MB of data we partition to 1 file per resource type. This gives fewer files for customers to download. 

When we have more than 100 MB of data we allow Glue to create as many files as needed per `resourceType`. With a DDB table that is several GB in size I've seen Glue makes a most of 19 files per `resourceType`. Each file is atleast 10 MB in size.

bulkExport/glueScripts/export-script.py
```diff
@@ -0,0 +1,124 @@
...
+    data_frame = data_source_cleaned_dyn_frame.toDF()
+    # If only a few records, we should combine the output to 1 file per resource
+    # For lots of record, combining all records into 1 file per resource causes memory out of limit error in Glue
+    # Each record is about 2kb, so 50,000 is about 100MB of data
+    if (total_records_to_export < 50000):
+        print('Combining data into fewer partitions')
+        data_frame = data_frame.coalesce(1)
+
+    # Create duplicated column so we can use it in partitionKey later
+    data_frame = data_frame.withColumn('resourceTypeDup', data_frame.resourceType)
```
nguyen102, line 82: When using `partitionKey` in line 95 it removes the field from the record, therefore I'm creating a duplicate of `resourceType` so that I can use `partitionKey` later on to split records into corresponding `resourceType` files. 

[link](https://aws.amazon.com/blogs/big-data/work-with-partitioned-data-in-aws-glue/)
"When you execute this write, the type field is removed from the individual records and is encoded in the directory structure. "

bulkExport/glueScripts/export-script.py
carvantes: I feel like this 100MB behavior is confusing for customers. I imagine that they'll try FWoA with test data and think that export returns one file per resource and only later when they run it in a prod-like environment they'll find out that it actually returns many files per resource.
Yes, the FHIR spec says that there can be multiple files per resource so it is compliant, but the 100MB threshold seems arbitrary. We could document this behavior but it sounds weird to me. 

Also, is there a way to set a maximum size for the S3 files that are written? That may be the ideal approach( idk if that's possible, I'm new to Spark)

nguyen102: After further research and testing, I was able to get the code working with `repartition(1)` instead of `coalesce(1)`. I tested it by exporting 4GB of data from DDB and the code worked as expected. Further details about `coalesce` vs `repartition` [here](https://stackoverflow.com/a/56263274)

nguyen102: Yep, setting a maximum size for S3 files written by Spark would be ideal. but unfortunately that's not possible. There are some hacky ways to do it by trying to approximate how much a file size would be given how many records you're writing to a file. However, that would cause the job to run slowly. 

This [link](https://stackoverflow.com/a/47229465) is about outputting the files to parquet, which is not exactly the same as json, but I believe the result will be the same.

carvantes (approved):

rsmayda (approved):

PR #131, opened 2020-10-30
feat: add bulk export state machine

carvantes:
Description of changes:

Add the state machine definition that orchestrates a bulk export job.

Screen Shot 2020-11-02 at 11 56 09

All Lambda tasks are using the default retry policy (MaxAttempts=3, IntervalSeconds=1, BackoffRate=2.0)

The `catchAllUpdateStatusToFailed` step is added to ensure that if ANY Lambda task fails, the export status is set to `failed`.
Using a Parallel state with a single branch is a trick to catch the error of any task inside it and redirect to the error handler. Otherwise we'd need to add a catch on each individual task and there'll be a lot of arrows pointing to the error handler and that's ugly.

The `serverless-step-functions` plugin facilitates working with step functions, but it compiles into raw CFN on serverless package/deploy so it won't cause us any trouble.

Checklist:


bulkExport/state-machine-definition.yaml
```diff
@@ -0,0 +1,120 @@
...
+                    StringEquals: 'FAILED'
+                  - Variable: "$.executionParameters.glueJobRunStatus"
+                    StringEquals: 'TIMEOUT'
+                  - Variable: "$.executionParameters.glueJobRunStatus"
+                    # STOPPING and STOPPED can only occur here if the job was forcefully stopped with a Glue API call from outside the FHIR server, so we treat it as failure
+                    StringEquals: 'STOPPING'
+                  - Variable: "$.executionParameters.glueJobRunStatus"
+                    StringEquals: 'STOPPED'
+                  Next: updateStatusToFailed
+            cancelExportJob:
```
nguyen102: In Glue the API that is used to "cancel" a job is called "stop". Let's use the same terminology as glue to be consistent and rename this to `stopExportJob`.

https://docs.aws.amazon.com/glue/latest/webapi/API_BatchStopJobRun.html

carvantes: done

nguyen102 (approved):

PR #135, opened 2020-10-30
fix: Dev deployment

rsmayda:
Description of changes:
- [dev deployment failing](https://github.com/awslabs/fhir-works-on-aws-deployment/runs/1332848654)
  - `  An error occurred: KibanaUserPool - Existing schema attributes cannot be modified or deleted. (Service: AWSCognitoIdentityProvider; Status Code: 400; Error Code: InvalidParameterException; Request ID: null; Proxy: null).`
- Fixing the Kibana user pool to be use the old SCHEMA 
- linted all yml files

Previous commit was deployed to a new environment instead of an existing one

Checklist:

### DIFF
--- a/cloudformation/elasticsearch.yaml
+++ b/cloudformation/elasticsearch.yaml
@@ -10,9 +10,6 @@ Resources:
     Properties:
       AutoVerifiedAttributes:
         - email
-      MfaConfiguration: 'OFF'
-      EmailVerificationSubject:
-        Ref: AWS::StackName
       UserPoolName: !Sub '${AWS::StackName}-Kibana'
       Schema:
         - Name: email
--- a/cloudformation/elasticsearch.yaml
+++ b/cloudformation/elasticsearch.yaml
@@ -10,18 +10,22 @@ Resources:
     Properties:
       AutoVerifiedAttributes:
         - email
+      MfaConfiguration: 'OFF'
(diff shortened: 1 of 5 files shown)


cloudformation/elasticsearch.yaml
```diff
@@ -15,16 +15,17 @@ Resources:
         Ref: AWS::StackName
       UserPoolName: !Sub '${AWS::StackName}-Kibana'
       Schema:
```
rsmayda, line 14: Now matching to be what it used to be: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/cloudformation/elasticsearch.yaml#L14-L19

cloudformation/elasticsearch.yaml
```diff
@@ -15,16 +15,17 @@ Resources:
         Ref: AWS::StackName
```
nguyen102: Do lines 13-15 above need to be reverted also?

rsmayda: When I tested it was fine. It is only the Schema variable https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-schema

nguyen102 (approved):

PR #137, opened 2020-11-02
fix: Bulk export glue script read directly from DDB

nguyen102:
Issue #, if available:

Description of changes:
* Read data from DDB directly instead of through Glue catalog. This runs much faster than before because it allows the Glue executors to work in parallel.
* We no longer need to run the Glue crawler and create Glue catalog tables first, because we're not reading DDB data through Glue catalog.
* Use `Dynamic Frame` instead of `Data Frame` to process and transform the data, because [DyFrame](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-crawler-pyspark-extensions-dynamic-frame.html) schemas are "self-describing". Using `Data Frame` will cause all records to have all fields. For example, an `Observation` record will have a `gender` field with the value of `null`. If we were to drop the `gender` field, then a `Patient` record will no longer have a `gender` field.
* Removed `crawler` definitions in `serverless.yaml` and `bulkExport.yaml`

Checklist:


bulkExport/glueScripts/export-script.py
```diff
@@ -37,13 +37,30 @@
...
-original_data_source_dyn_frame = glueContext.create_dynamic_frame.from_catalog(database = glue_database, table_name = glue_table_name)
+# dynamodb.splits is determined by the formula from the weblink below
+# https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-connect.html#aws-glue-programming-etl-connect-dynamodb
+if (worker_type != "G.2X" and worker_type != "G.1X"):
+    raise Exception(f"Worker type {worker_type} not supported. Please choose either worker G2.X or G1.X")
+
+num_executors = int(number_workers) - 1
+num_slots_per_executor = 8
+if worker_type == 'G.2X':
+    num_slots_per_executor = 16
```
carvantes: Do not mutate `num_slots_per_executor`.

Here you can just assign once using a ternary.

cloudformation/bulkExport.yaml
```diff
@@ -197,8 +100,8 @@ Resources:
     Properties:
       Role: !GetAtt GlueJobRole.Arn
       GlueVersion: '2.0'
-      WorkerType: 'G.2X'
-      NumberOfWorkers: 5
+      WorkerType: '${self:custom.exportGlueWorkerType}'
+      NumberOfWorkers: '${self:custom.exportGlueNumberWorkers}'
```
carvantes: use raw CFN params instead of serverless custom options.

Or maybe just keep this options hardcoded(right now they are hardcoded in `serverless.yaml` anyways).
Do we want to support customization for these 2 values? I don't think It's a must for the 1st release.

carvantes (changes requested):

cloudformation/bulkExport.yaml
nguyen102: Although not strictly required for the first release, these two values are something that customer will want to play around with for their setup. Customers will want to adjust these values to fit their needs. Customers with large DDB tables might want to increase the number of workers to get faster export speed at a higher price. Customers with smaller DDB tables might want less workers and a different worker type to fit their needs.

nguyen102: Updated to use CFN params.

bulkExport/glueScripts/export-script.py
nguyen102: Updated

carvantes (approved):

PR #138, opened 2020-11-03
fix: add missing step functions permissions

carvantes:
Description of changes:
Fix env variable and add permissions. 
Also created a separate `UpdateStatusLambdaRole` to break a circular dependency( and because it was the right thing to do to begin with)

btw, the `iamRoleStatements` in the `provider` section are added to ALL lambdas unless they provide their own role. That's probably not what we want since those permissions are specific to the `fhirServer` lambda. Still, not changing it right now since that refactor would cause even more conflicts with mainline.

Checklist:

### DIFF
--- a/cloudformation/bulkExport.yaml
+++ b/cloudformation/bulkExport.yaml
@@ -138,3 +138,32 @@ Resources:
                 Action:
                   - s3:GetObject
                 Resource: !Join ['', [!GetAtt BulkExportResultsBucket.Arn, '/*']]
+
+  UpdateStatusLambdaRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service: 'lambda.amazonaws.com'
+            Action: 'sts:AssumeRole'
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+        - arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess
+      Policies:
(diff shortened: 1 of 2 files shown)


cloudformation/bulkExport.yaml
```diff
@@ -138,3 +138,32 @@ Resources:
...
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - dynamodb:UpdateItem
+                Resource:
+                  - !GetAtt ExportRequestDynamoDBTable.Arn
+              - Effect: Allow
+                Action:
+                  - kms:Decrypt
```
rsmayda, line 167: When do we need KMS decryption for a DDB update? I would think encryption is needed?

rsmayda (commented):
Good callout:

> btw, the iamRoleStatements in the provider section are added to ALL lambdas unless they provide their own role. That's probably not what we want since those permissions are specific to the fhirServer lambda.

Can you add this as tech debt to our internal tracking system?

cloudformation/bulkExport.yaml
carvantes: kms:Decrypt is correct. 

You can find the details here: https://docs.aws.amazon.com/kms/latest/developerguide/services-dynamodb.html

tl;dr This allow Is not about decrypting or encrypting DDB table contents. This permissions is needed to decrypt the "table key". DynamoDB internally manages the encryption/decryption of your data.

carvantes:
> Can you add this as tech debt to our internal tracking system?

done

cloudformation/bulkExport.yaml
nguyen102: @carvantes  thanks for the clarification.

nguyen102 (approved):

PR #139, opened 2020-11-05
feat: upload glue scripts to s3 on deployment

carvantes:
Description of changes:
Add a custom resource to upload glue scripts to s3 as part of the CFN deployment. 

Checklist:

### DIFF
--- a/bulkExport/uploadGlueScriptsToS3.ts
+++ b/bulkExport/uploadGlueScriptsToS3.ts
@@ -1,0 +1,64 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs';
+import * as AWS from 'aws-sdk';
+import axios from 'axios';
+
+const sendCfnResponse = async (event, status: 'SUCCESS' | 'FAILED', error?: Error) => {
+    const responseBody = JSON.stringify({
+        Status: status,
+        Reason: error?.message,
+        // The value of PhysicalResourceId doesn't really matter in this case.
+        // It just needs to be the same string on all responses to indicate that it is the same resource.
+        PhysicalResourceId: 'glueScripts',
+        StackId: event.StackId,
+        RequestId: event.RequestId,
+        LogicalResourceId: event.LogicalResourceId,
+    });
+    console.log(`Sending response to CFN: ${responseBody}`);
+    await axios.put(event.ResponseURL, responseBody);
+};
+/**
+ * Custom resource lambda handler that uploads a specific file to s3.
+ * Custom resource spec: See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html
+ * @param event Custom resource request event. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requests.html
+ */
+exports.handler = async event => {
+    console.log(event);
+    try {
+        if (process.env.GLUE_SCRIPTS_BUCKET === undefined) {
+            throw new Error('Missing env variable GLUE_SCRIPTS_BUCKET');
+        }
+        const s3 = new AWS.S3();
+        const filename = 'export-script.py';
+        const path = `bulkExport/glueScripts/${filename}`;
+
+        if (event.RequestType === 'Create' || event.RequestType === 'Update') {
+            console.log(`uploading ${filename} to ${process.env.GLUE_SCRIPTS_BUCKET}`);
+            await s3
+                .putObject({
+                    Bucket: process.env.GLUE_SCRIPTS_BUCKET,
+                    Body: fs.readFileSync(path),
+                    Key: filename,
+                })
+                .promise();
+            console.log('upload successful');
+            await sendCfnResponse(event, 'SUCCESS');
+        } else {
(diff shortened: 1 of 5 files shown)


serverless.yaml
```diff
@@ -215,6 +219,17 @@ functions:
...
     handler: bulkExport/index.updateStatusStatusHandler
 
+  uploadGlueScripts:
+    timeout: 30
+    memorySize: 192
+    runtime: nodejs12.x
+    description: 'Upload glue scripts to s3'
+    role: UploadGlueScriptsLambdaRole
+    handler: bulkExport/uploadGlueScriptsToS3.handler
+    disableLogs: true # needed to avoid race condition error "Resource of type 'AWS::Logs::LogGroup' already exists" since the custom resource lambda invocation may create the log group before CFN does
```
rsmayda, line 229: Interesting - does this mean we will not have logs for the UploadGlueScriptsLambda? If we still do console.log() in the lambda does that just write locally and we are unable to see it?

bulkExport/uploadGlueScriptsToS3.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+    console.log(`Sending response to CFN: ${responseBody}`);
+    await axios.put(event.ResponseURL, responseBody);
+};
+/**
+ * Custom resource lambda handler that uploads a specific file to s3.
+ * Custom resource spec: See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html
+ * @param event Custom resource request event. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requests.html
+ */
+exports.handler = async event => {
```
rsmayda, line 29: So this lambda is kicked off by the CFN custom resource deployment, and we ensure that the lambda is invoked every deployment by having that RandomValue parameter inside the custom resource.

Does the deployment order matter? ie CustomResource deployed first _then_ this lambda?

bulkExport/uploadGlueScriptsToS3.ts
```diff
@@ -0,0 +1,64 @@
...
+    console.log(event);
+    try {
+        if (process.env.GLUE_SCRIPTS_BUCKET === undefined) {
+            throw new Error('Missing env variable GLUE_SCRIPTS_BUCKET');
```
rsmayda, line 33: If this error throws - will the deployment fail?

Could be race condition that the glue script was never uploaded to S3?

rsmayda (dismissed):

bulkExport/uploadGlueScriptsToS3.ts
carvantes: Custom resources are only updated when their parameters change(similar to regular CFN resources). The random value forces the update to happen on all deployments (there's no harm in re-uploading the same file)

A custom resource is just like any other CFN resource. CFN determines the deployment order from the dependency graph. The custom resource implicitly depends on the lambda function since it references it at:
```
ServiceToken: !GetAtt UploadGlueScriptsLambdaFunction.Arn
```

serverless.yaml
carvantes: No, it just means that Serverless won't explicitly create the LogGroup by generating an `AWS::Logs::LogGroup` resource. 

Lambda functions create the corresponding LogGroup/LogStream when they are invoked for the first time if it does not exist already.

We'll have Lambda logs as usual.

bulkExport/uploadGlueScriptsToS3.ts
rsmayda: Got it thanks - should the AWS::Glue::Job resource depend on the custom resource (to help ensure that the file has been uploaded to S3)?

bulkExport/uploadGlueScriptsToS3.ts
carvantes: yes, deployment will fail. that's what we want, there's no way to recover. The CFN template is wrong in that case.

No. This custom resource only succeeds after the upload to s3 is complete. CFN will wait until it gets a `SUCCESS` response from the custom resource before completing the CFN deployment.

rsmayda (approved):

bulkExport/uploadGlueScriptsToS3.ts
carvantes: I added the `DependsOn`.

Most of the time it doesn't matter since the python script is just needed at run time, not at deploy time. But it does prevent the very rare case where a customer manages to start a bulk export request BEFORE their first CREATE stack operation completes and the job fails to find the script in s3.

serverless.yaml
```diff
@@ -5,6 +5,10 @@
 
 service: fhir-service
 
+package:
+  include:
+    - bulkExport/glueScripts/export-script.py
```
nguyen102, line 10: It might be worth moving `bulkExport/glueScripts/export-script.py` to `bulkExport/export-script.py`. I think we put the script in a folder before because we thought we have to upload everything in a folder to S3. However, if we can select specific files, putting the script in it's own folder seems redundant.

nguyen102 (approved):
Looks good. Just one minor comment about moving `export-script.py` one directory up. It's not a merge blocking change though.

serverless.yaml
carvantes: I'll keep it as is. There's no harm in having an extra folder, plus I like how it separates the `.py` files from the rest.

PR #140, opened 2020-11-06
feat: (Part 1) Bulk export integration tests

nguyen102:
Issue #, if available:

Description of changes:
This PR adds Integration Tests for the Export feature. This PR adds three tests
* Successfully exporting all DB data after a set time
* Successfully exporting DB data specified by a "type"
* Successfully stop an export job that is currently running

I wanted to release this PR now and get feedback while I work on Part 2 of Integration tests. Part 2 will revolve around adding these tests to our CI/CD pipeline. Like the Crucible tests, these tests will run after a developer merges to the `develop` branch. 

Checklist:


package.json
```diff
@@ -78,6 +82,8 @@
     "transform": {
       "\\.(ts)$": "ts-jest"
     },
+    "testEnvironment": "node",
```
nguyen102, line 88: Jest normally runs in "jsdom", which is a browser like environment. Since this is a node app, we should run the tests in "node".
https://jestjs.io/docs/en/configuration

package.json
```diff
@@ -78,6 +82,8 @@
...
+    "testRunner": "jest-circus/runner",
```
nguyen102, line 89: By default, if the `beforeAll` hook throws an error during the test, jest will not automatically fail the test.

This config option fixes that.
https://github.com/facebook/jest/issues/2713

integration-tests/bulkExport.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    );
+
+    function checkOnlyFilesSpecifiedByTypeAreExported(outputs: ExportStatusOutput[], type: string) {
+        const expectedFileTypes = type.split(',');
+        const fileTypesExported = uniq(
+            outputs.map(out => {
+                return out.type;
+            }),
+        );
+        expect(isEqual(expectedFileTypes.sort(), fileTypesExported.sort())).toBeTruthy();
```
carvantes: Use `.toEqual` from jest instead. 

The big difference is that if the test fails you get to see the object diff instead of just a "expected truthy but got false"

https://jestjs.io/docs/en/expect#toequalvalue

integration-tests/bulkExport.test.ts
```diff
@@ -0,0 +1,168 @@
...
+describe('Bulk Export', () => {
+    let bulkExportTestHelper: BulkExportTestHelper;
+
+    beforeAll(() => {
+        if (API_URL === undefined) {
+            throw new Error('API_URL environment variable is not defined');
+        }
+        if (API_ACCESS_TOKEN === undefined) {
+            throw new Error('API_ACCESS_TOKEN environment variable is not defined');
+        }
```
carvantes: Where is the `API_ACCESS_TOKEN ` coming from? The token expires, so it cannot be a fixed secret env variable like we do for AWS access key in the other integ tests.

It may be easier to have the integ tests call cognito to get an access token.

integration-tests/bulkExport.test.ts
```diff
@@ -0,0 +1,168 @@
...
+        },
+        FIVE_MINUTES_IN_MS,
+    );
+
+    function checkOnlyFilesSpecifiedByTypeAreExported(outputs: ExportStatusOutput[], type: string) {
+        const expectedFileTypes = type.split(',');
+        const fileTypesExported = uniq(
+            outputs.map(out => {
+                return out.type;
+            }),
```
carvantes: nit: I think one liners are more readable if you omit the `return` word
```suggestion
            outputs.map(out => out.type),
```

carvantes (changes requested):

integration-tests/bulkExportTestHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+            if (startExportJobParam.since) {
+                // eslint-disable-next-line no-underscore-dangle
+                params._since = startExportJobParam.since.toISOString();
+            }
+            if (startExportJobParam.type) {
+                // eslint-disable-next-line no-underscore-dangle
+                params._type = startExportJobParam.type;
+            }
+            const urlSearchParams = new URLSearchParams(params);
+            const response = await this.fhirUserAxios.get(`/$export?${urlSearchParams.toString()}`);
```
carvantes: axios can take `params`. There's no need to manually build the url params string
```js
axios.get('/user?ID=12345')
// same as
axios.get('/user', {
    params: {
      ID: 12345
    }
  })
```

integration-tests/bulkExport.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            return checkResourceInExportedFiles(
+                responseBody.output,
+                resTypToResExpectedInExport,
+                resTypToResNotExpectedInExport,
+            );
+        },
+        FIVE_MINUTES_IN_MS,
+    );
+
+    function checkOnlyFilesSpecifiedByTypeAreExported(outputs: ExportStatusOutput[], type: string) {
```
carvantes: I don't like how this function is declared between tests. It's better to just keep `test` declarations within the `describe` block.

Also, this function is only used once. I thin you can just inline on L150 like 
```js
expect(new Set((responseBody.output as ExportStatusOutput[]).map(x => x.type))).toEqual(new Set([type]));
```

integration-tests/bulkExport.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            res.id = id;
+            res.meta = {
+                lastUpdated: entry.response.lastModified,
+                versionId: entry.response.etag,
+            };
+            resourceTypeToExpectedResource[resourceType] = res;
+        });
+
+        return resourceTypeToExpectedResource;
+    };
```
carvantes: L50 is mutating the Bundle object that comes from the JSON file. That Bundle is shared by all tests and should NOT be mutated. Looks like that's not causing any trouble now but it's brittle.

Also, I think that this function could be simplified. I think that we could use the fact that the bundle response preserves the order of the resources. Just loop through the response and copy the id to the bundle item with the same index.

The way it's written now is a bit hard to follow and breaks if there are more than one resource with the same type

integration-tests/bulkExport.test.ts
nguyen102: updated

integration-tests/bulkExport.test.ts
nguyen102: updated

integration-tests/bulkExportTestHelper.ts
nguyen102: updated

integration-tests/bulkExport.test.ts
nguyen102: updated to use this new one line transform instead of the function

integration-tests/bulkExport.test.ts
nguyen102: The `accessToken` is provided as an environment variable. For CI/CD with github actions we generate it on the command line and pass it into the test. However, for running this locally, I think it would be more convenient if we don't have to manually generate the token each time. 

I'll add Cognito so that the access token can be auto generated.

integration-tests/bulkExport.test.ts
nguyen102: Good call about L50, I added `cloneDeep` from `lodash` so the modifications to `createBundle` is to a local copy, and not globally.

I simplified the method to use the `bundleResponse` order in creating the new `resourceTypeToExpectedResource` variable. And yep, you're right about how this data structure only supports one value for each `resourceType`. I wrote the test this way to keep the test simpler. We could make value of the `resourceTypeToExpectedResource` object be an array of resources. However, the test doesn't require us checking for multiple resources per `resourceType`, therefore an array is not necessary.

carvantes (approved):

PR #148, opened 2020-11-10
fix: add root path

rsmayda:
Description of changes:
- Fix root path & failing build: https://github.com/awslabs/fhir-works-on-aws-deployment/runs/1380661219

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -95,7 +95,7 @@ functions:
   fhirServer:
     timeout: 40
     description: 'FHIR API Server'
-    events: 
+    events:
       - http:
           authorizer:
             type: COGNITO_USER_POOLS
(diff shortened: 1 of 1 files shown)


serverless.yaml
```diff
@@ -106,7 +106,19 @@ functions:
               - 'profile'
               - 'aws.cognito.signin.user.admin'
           method: ANY
-          path: '/{proxy+}'
+          path: '/'
```
nguyen102: I'm not sure if this will make a difference, but when testing, I didn't require quotes around `/`. 

It just looked like this.
`path: /`

https://www.serverless.com/framework/docs/providers/aws/events/apigateway/

nguyen102 (dismissed):
Optional comment about quotes.

nguyen102 (approved):

serverless.yaml
carvantes: @nguyen102, that's just YAML syntax that allows unquoted strings. The following are equivalent:

yaml
```yaml
string1: /
string2: '/'
string3: "/"

```
json
```json
{
  "string1": "/",
  "string2": "/",
  "string3": "/"
}
```

PR #170, opened 2020-12-22
fix: formatting and make seperate roles per lambda

rsmayda:
Description of changes:
- cosmetic: change from using `Ref:{}` to using `!Ref`
- cosmetic: Update MD files to use syntax highlighting
- fix: Update Install wording
- fix: Removed iamRoleStatements in favor or more specific FhirServerLambdaRole
- fix: Changed ES permissions from `*` to `ESHttp*`
- fix: Changed S3 permissions from `*` to `s3:*Object`; `s3:ListBucket`; `s3:DeleteObjectVersion`

Checklist:

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -21,7 +21,8 @@ This installation guide covers a basic installation on Windows, Unix-like system
 In a Terminal application or command shell, navigate to the directory containing the package’s code.
 
 Configure your AWS Credentials:
-```
+
+```sh
 aws configure
 ```
 
@@ -45,7 +46,7 @@ You can also use their abbreviations:
 sudo ./scripts/install.sh -r <REGION> -s <STAGE>
 ```
 
-#### Troubleshooting
+#### Linux or OSX Troubleshooting
 
 If your PATH or environment variables are not accessible to the root/sudo user, you can try to use this command:
 
@@ -62,6 +63,7 @@ sudo -E ./scripts/install.sh
 ```
 
 If you are using AWS Named Profiles, please use the `AWS_PROFILE` environment variable to set it, and use the `-E` sudo flag :
+
 ```sh
 export AWS_PROFILE=myprofile
 sudo -E ./scripts/install.sh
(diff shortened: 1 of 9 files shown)


serverless.yaml
```diff
@@ -270,6 +195,7 @@ resources:
         DeletionPolicy: Retain
         UpdateReplacePolicy: Retain
         Properties:
+          TableName: '${self:custom.resourceTableName}'
```
carvantes, line 198: Changing the TableName requires replacement of the ddb table. Is that what we want?

** I'm sure it requires replacement when updating the name. I'm not 100% sure about what happens when changing from some table name to not specifying any name, but I'd expect it to be replaced.

rsmayda: Good call out I am actually just moving it from below in the properties; no functional change here. Old line was 284

serverless.yaml
```diff
@@ -294,10 +219,11 @@ resources:
                 Value: 'daily'
               - Key: 'service'
                 Value: 'fhir'
-            - Ref: AWS::NoValue
+            - !Ref AWS::NoValue
       ExportRequestDynamoDBTable:
         Type: AWS::DynamoDB::Table
         Properties:
+          TableName: '${self:custom.exportRequestTableName}'
```
carvantes, line 226: Same thing here about renaming ddb tables

carvantes (dismissed):

serverless.yaml
carvantes: Right. My bad

INSTALL.md
```diff
@@ -45,7 +46,7 @@ You can also use their abbreviations:
 sudo ./scripts/install.sh -r <REGION> -s <STAGE>
 ```
 
-#### Troubleshooting
+#### Linux or OSX Troubleshooting
```
nguyen102: Nitpick: I think OSX is just called macOS now.
https://en.wikipedia.org/wiki/OSX_(disambiguation)
`OS X was until September 2016 the name of an operating system family for Apple computers, thereafter known as macOS. `

nguyen102 (dismissed):
Overall looks good. One small nitpick.

INSTALL.md
rsmayda: a little TIL!

nguyen102 (dismissed):

nguyen102 (approved):

PR #173, opened 2020-12-31
chore: bump jest and typescript version

carvantes:
Description of changes:

upgrade jest to patch a vulnerability in node-notifier
https://github.com/advisories/GHSA-5fw9-fq32-wv5p

Also updated typescript to the latest version. One thing led to another. the latest jest types use features of typescript introduced in the latest versions. the bump to typescript is a major version bump but it is safe since it is a dev dependency with no impact at runtime(as long as it compiles it's fine). Also, the breaking changes are mostly around browser support.


auditLogMover/package.json
```diff
@@ -5,23 +5,24 @@
...
+    "jest": "^26.6.3",
+    "serverless": "^1.72.0",
+    "serverless-plugin-typescript": "^1.1.9",
+    "serverless-step-functions": "^2.19.0",
+    "sinon": "^9.0.2",
+    "ts-jest": "^26.4.4",
+    "typescript": "^4.1.3"
   },
   "dependencies": {
+    "@types/jest": "^26.0.19",
```
rsmayda: Needed in dependencies?

carvantes: no. I just moved it to dev dependencies.

Turns out that `yarn upgrade` also installs packages if not already installed

rsmayda (approved):

PR #176, opened 2021-01-05
chore: Update dockerfile used for building docker image

nguyen102:
Issue #, if available:

Description of changes:
* Removed installing `dnsutil` as that is not needed
* `less` was not removed as it's required by AWS CLI. [More info here](https://github.com/aws/aws-cli/issues/5038)
*  Added `RUN rm -rf ./fhir-works-on-aws-deployment/.build`. If a user runs `sls` locally before building the docker image, the `.build` folder needs to be removed. 

I tested building the new docker image and using the image to deploy FHIR Works.

Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:14.4
 
 # Required dependencies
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /home/node
 # Temporary use local copy
 RUN mkdir fhir-works-on-aws-deployment
 COPY ./ ./fhir-works-on-aws-deployment/
+RUN rm -rf ./fhir-works-on-aws-deployment/.build
```
rsmayda, line 13: Just to confirm if there isn't any .Build directories. This command does not error out￼ and fails the build

nguyen102: Confirmed. `rm -rf` removes the directory if it's there, but doesn't error out if the directory isn't there.

rsmayda (approved):

PR #181, opened 2021-01-11
feat: support standard FHIR search parameters

carvantes:
This PR adds the following features:
- FHIR search parameters can now be used in queries (i.e, the ones at the bottom of https://www.hl7.org/fhir/patient.html). Each FHIR resource type has a different set of search parameters. The complete list of search parameters can be found on the FHIR documentation.
- The capability statement returned from `/metadata` now includes the details of all search parameters supported by FWoA.
- Search requests using invalid search parameters now return an error instead of an empty result set

Most of the changes are found in the interface, search and routing packages. Here we are just bumping package versions and adding integration tests.

**Testing:** 
I deployed these changes and ran all tests successfully on my AWS account.

**Notes:**
- updated bulk export to filter out the `_references` field introduced by https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/44

**Checklist:**

### DIFF
--- a/integration-tests/search.test.ts
+++ b/integration-tests/search.test.ts
@@ -1,0 +1,87 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import { AxiosInstance } from 'axios';
+import waitForExpect from 'wait-for-expect';
+import { getFhirClient, randomPatient } from './utils';
+
+jest.setTimeout(60 * 1000);
+
+const expectResourceToBePartOfSearchResults = async (
+    client: AxiosInstance,
+    search: { url: string; params?: any },
+    resource: any,
+) => {
+    console.log('Searching with params:', search);
+    await expect(
+        (async () => {
+            return (
+                await client.get(search.url, {
+                    params: search.params,
+                })
+            ).data;
+        })(),
+    ).resolves.toMatchObject({
+        resourceType: 'Bundle',
+        entry: expect.arrayContaining([
+            expect.objectContaining({
+                resource,
+            }),
+        ]),
+    });
+};
+
+describe('search', () => {
+    let client: AxiosInstance;
+    beforeAll(async () => {
+        client = await getFhirClient();
+    });
+    test('search for various valid parameters', async () => {
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', randomPatient())).data;
+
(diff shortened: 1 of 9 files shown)


integration-tests/search.test.ts
```diff
@@ -0,0 +1,88 @@
...
+            p({ gender: testPatient.gender, _id: testPatient.id }), // gender alone is not unique enough to guarantee a match on the first page of search results
+            p({ phone: testPatient.telecom.find(x => x.system === 'phone')!.value }),
+            p({ email: testPatient.telecom.find(x => x.system === 'email')!.value }),
+            p({ telecom: testPatient.telecom.find(x => x.system === 'email')!.value }),
+            p({ organization: testPatient.managingOrganization.reference }),
+        ];
+
+        // run tests serially for easier debugging and to avoid throttling
+        // eslint-disable-next-line no-restricted-syntax
+        for (const testParams of testsParams) {
```
nguyen102, line 77: Nitpick: `testParams` and `testsParams` are very similar looking and it's easy to confuse the two.

We could make `testParams` something as small as `tp` to differentiate the two.

integration-tests/search.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    let client: AxiosInstance;
+    beforeAll(async () => {
+        client = await getFhirClient();
+    });
+    test('search for various valid parameters', async () => {
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', randomPatient())).data;
+
+        // wait for the patient to be asynchronously written to ES
+        await waitForExpect(
+            expectResourceToBePartOfSearchResults.bind(
```
nguyen102, line 45: what does `bind` do here?

nguyen102 (approved):
Approved with one nitpick non blocking comment.

integration-tests/search.test.ts
carvantes: I'm using it here since `waitForExpect` expects a function as a parameter and I'm "setting" the arguments for expectResourceToBePartOfSearchResults.

If you are not familiar with `.bind` I suggest you read the docs. It is a very powerful/useful function.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

integration-tests/search.test.ts
carvantes: makes sense. Still, the var names are understandable on their own, and the code inside the for loop is just one line so there's not much room for confusion. I'll keep it as is.

PR #183, opened 2021-01-13
chore: Add @types/cors to dev dependency

nguyen102:
Issue #, if available:

Description of changes:
This resolves a warning when running `sls deploy`
```
/Volumes/unix/workplace/fhir-works-on-aws-deployment/node_modules/fhir-works-on-aws-routing/lib/app.d.ts (2,29): Could not find a declaration file for module 'cors'. '/Volumes/unix/workplace/fhir-works-on-aws-deployment/node_modules/cors/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/cors` if it exists or add a new declaration (.d.ts) file containing `declare module 'cors';`
```

Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "serverless-http": "^2.3.1"
   },
   "devDependencies": {
+    "@types/cors": "^2.8.9",
     "@types/express": "^4.17.2",
     "@types/jest": "^26.0.19",
     "@types/node": "^12",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1277,6 +1277,11 @@
   dependencies:
     "@types/node" "*"
 
+"@types/cors@^2.8.9":
+  version "2.8.9"
+  resolved "https://registry.yarnpkg.com/@types/cors/-/cors-2.8.9.tgz#4bd1fcac72eca8d5bec93e76c7fdcbdc1bc2cd4a"
+  integrity sha512-zurD1ibz21BRlAOIKP8yhrxlqKx6L9VCwkB5kMiP6nZAhoF5MvC7qS1qPA7nRcr1GJolfkQC7/EAL4hdYejLtg==
+
 "@types/express-serve-static-core@*":
   version "4.17.9"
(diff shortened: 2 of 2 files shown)


carvantes (approved):
The fact that this bug is on mainline means that our gh actions are not set up properly. They are not doing a proper yarn run release. Can you fix that?

carvantes (changes requested):
nvm my previous approval. This doesn't look right.

This package does not depend on cors directly. The routing package is the one that needs to include the cors types

rsmayda:
> nvm my previous approval. This doesn't look right.
> 
> This package does not depend on cors directly. The routing package is the one that needs to include the cors types

It looks like routing does - https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/package.json#L45

This may actually be an sls issue/concern on how they bundle it up?

Looking at app.ts it looks like this is a param that the deployment package must pass in: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/app.ts#L30

@nguyen102 Are you getting this error when running `yarn run release` or just when you are running `sls deploy`?

carvantes:
> > nvm my previous approval. This doesn't look right.
> > This package does not depend on cors directly. The routing package is the one that needs to include the cors types
> 
> It looks like routing does - https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/package.json#L45
> 
> This may actually be an sls issue/concern on how they bundle it up?
> 
> Looking at app.ts it looks like this is a param that the deployment package must pass in: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/app.ts#L30
> 
> @nguyen102 Are you getting this error when running `yarn run release` or just when you are running `sls deploy`?

The routing package declares it on `devDependencies` but it should declare it in `dependencies`. This is because it is using the cors types as part of its own exported type declarations (the type of the function param you mentioned)

rsmayda:
@carvantes Thanks for the explainer and it makes sense -- closing this PR in favor of the routing one: https://github.com/awslabs/fhir-works-on-aws-routing/pull/43

PR #185, opened 2021-01-14
chore: Update routing dependency to version 4.0.1

nguyen102:
Issue #, if available:

Description of changes:
chore: Update routing dependency to version 4.0.1

Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "fhir-works-on-aws-authz-rbac": "4.1.0",
     "fhir-works-on-aws-interface": "7.0.0",
     "fhir-works-on-aws-persistence-ddb": "3.0.0",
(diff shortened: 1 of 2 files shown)


carvantes:
This fixes the immediate issue, but I think we should also update the gh action to prevent this kind of bugs from reaching mainline again. It should run to run the full `yarn run release` instead of just lint & test

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/.github/workflows/unit-test.yaml#L30-L41

nguyen102:
@carvantes yep, agreed. I'm going to add that to the GH action in a separate PR. I want to keep the PRs separate so when we do "Squash and merge" through the UI it doesn't combine the commits into one.

carvantes (approved):

PR #192, opened 2021-01-20
feat: add Implementation Guides compiler script 

justinusmenzel:
Issue #, if available: INT-931

Description of changes:
Added a IGCompiler TypeScript class to src and a compile-igs.ts script to the scripts folder.  
Its purpose of the class is 
 * to search through a directory and find any sub directory containing an IG package.
 * collect high level name and version info for each IG package
 * make sure there aren't any missing dependencies 
 * make sure there are no circular dependencies
 * extract all the SearchParameters (needs to be parameterized in the future)
 * pass list of SearchParameter instances to SearchImplementationGuides.compile function
 * store output in a JSON file

The compile-igs.ts script does
 * parse command line arguments
 * import ImplementationGuides implementation (implementations in the future
 * invoke IGCompiler
 * catch and report errors

Checklist:

### DIFF
--- a/src/IGCompiler.test.ts
+++ b/src/IGCompiler.test.ts
@@ -1,0 +1,179 @@
+/* eslint no-restricted-syntax: 0 */
+/* eslint no-await-in-loop: 0 */
+/* eslint class-methods-use-this: 0 */
+
+import { dir, DirectoryResult } from 'tmp-promise';
+import { join } from 'path';
+import { existsSync, mkdirSync, PathLike } from 'fs';
+import { ImplementationGuides } from 'fhir-works-on-aws-interface';
+import { IGCompiler, IGCompilerOptions, loadJson, storeJson } from './IGCompiler';
+
+class MockImplementationGuides implements ImplementationGuides {
+    compile(input: any[]): Promise<any> {
+        return Promise.resolve({
+            input,
+        });
+    }
+}
+
+interface IGVersion {
+    version: string;
+    deps: string[];
+}
+
+describe('IGCompiler tests', function() {
+    let workDir: DirectoryResult;
+    let igsDir: PathLike;
+    let output: string;
+
+    async function createIGs(options: IGCompilerOptions, igs: { [key: string]: IGVersion }) {
+        for (const [igName, igInfo] of Object.entries(igs)) {
+            const igPath = join(igsDir.toString(), igName);
+            mkdirSync(igPath);
+            await storeJson(join(igPath, 'searchParam1.json'), { igName, name: 'param1' });
+            await storeJson(join(igPath, 'searchParam2.json'), { igName, name: 'param2' });
+            const indexJson = {
+                files: [
+                    {
+                        resourceType: 'SearchParameter',
+                        filename: 'searchParam1.json',
+                    },
+                    {
+                        resourceType: 'SearchParameter',
+                        filename: 'searchParam2.json',
+                    },
+                    {
+                        resourceType: 'ValueSet',
+                        filename: 'searchParam1.json',
+                    },
+                ],
+            };
+            await storeJson(join(igPath, '.index.json'), indexJson);
+            const dependencies: { [key: string]: string } = {};
+            igInfo.deps.forEach((value: string) => {
+                const { version } = igs[value];
+                dependencies[value] = version;
+            });
+            await storeJson(join(igPath, 'package.json'), {
+                name: igName,
+                version: igInfo.version,
+                url: `http://http://hl7.org/fhir/${igName}`,
+                dependencies,
+            });
+        }
+        const implementationGuides = new MockImplementationGuides();
+        const igCompiler = new IGCompiler(implementationGuides, options);
+        console.log('Done creating IGs');
+        return igCompiler;
+    }
+
+    beforeEach(async function() {
+        workDir = await dir({ unsafeCleanup: true });
+        igsDir = join(workDir.path, 'igs');
+        mkdirSync(igsDir);
+        output = join(workDir.path, 'output.json');
+        console.log('before each');
+    });
+
+    afterEach(async function() {
+        console.log('cleaning up');
+        await workDir.cleanup();
+    });
+
+    it('compile a few IGs', async function() {
+        const options: IGCompilerOptions = {
+            ignoreVersion: true,
+        };
+        const igCompiler = await createIGs(options, {
+            'hl7.fhir.r4.core': {
+                version: '4.0.1',
+                deps: [],
+            },
+            'hl7.fhir.us.carin-bb': {
+                version: '1.0.0',
+                deps: ['hl7.fhir.r4.core', 'hl7.fhir.us.core'],
+            },
+            'hl7.fhir.us.core': {
+                version: '3.1.0',
+                deps: ['hl7.fhir.r4.core'],
+            },
+            'hl7.fhir.us.davinci-pdex-plan-net': {
+                version: '1.0.0',
+                deps: ['hl7.fhir.us.carin-bb'],
+            },
+            'us.nlm.vsac': {
+                version: '0.3.0',
+                deps: ['hl7.fhir.us.core'],
+            },
+        });
+        await igCompiler.compileIGs(igsDir, output);
+        expect(existsSync(output)).toEqual(true);
+        expect(await loadJson(output)).toEqual({
+            input: [
+                {
+                    igName: 'hl7.fhir.us.carin-bb',
+                    name: 'param1',
+                },
+                {
+                    igName: 'hl7.fhir.us.carin-bb',
+                    name: 'param2',
+                },
+                {
+                    igName: 'hl7.fhir.us.core',
+                    name: 'param1',
+                },
+                {
+                    igName: 'hl7.fhir.us.core',
+                    name: 'param2',
+                },
+                {
+                    igName: 'hl7.fhir.us.davinci-pdex-plan-net',
+                    name: 'param1',
+                },
+                {
+                    igName: 'hl7.fhir.us.davinci-pdex-plan-net',
+                    name: 'param2',
+                },
+                {
+                    igName: 'us.nlm.vsac',
+                    name: 'param1',
+                },
+                {
+                    igName: 'us.nlm.vsac',
+                    name: 'param2',
+                },
+            ],
+        });
+    });
+
+    it('circular dependencies', async function() {
+        const options: IGCompilerOptions = {
+            ignoreVersion: false,
+        };
+        const igCompiler = await createIGs(options, {
+            'hl7.fhir.r4.core': {
+                version: '4.0.1',
+                deps: [],
+            },
+            'hl7.fhir.us.carin-bb': {
+                version: '1.0.0',
+                deps: ['hl7.fhir.r4.core', 'hl7.fhir.us.core'],
+            },
+            'hl7.fhir.us.core': {
+                version: '3.1.0',
+                deps: ['hl7.fhir.r4.core', 'us.nlm.vsac'],
+            },
+            'hl7.fhir.us.davinci-pdex-plan-net': {
+                version: '1.0.0',
+                deps: ['hl7.fhir.us.carin-bb'],
+            },
+            'us.nlm.vsac': {
+                version: '0.3.0',
+                deps: ['hl7.fhir.us.core', 'hl7.fhir.us.davinci-pdex-plan-net'],
+            },
+        });
+        await expect(igCompiler.compileIGs(igsDir, output)).rejects.toThrow(
+            'Circular dependency found: hl7.fhir.us.carin-bb@1.0.0 -> hl7.fhir.us.core@3.1.0 -> us.nlm.vsac@0.3.0 -> hl7.fhir.us.core@3.1.0',
+        );
+    });
+});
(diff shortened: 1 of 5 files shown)


scripts/compile-igs.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+    if (igPath === null) {
+        console.error('Need path to IGs folder as argument');
+        usage();
+        process.exit(1);
+    }
+    const options = {
+        ignoreVersion,
+    };
+    return new IGsCompileHelper(options).compileIGs(igPath);
```
carvantes: Errors should be handled here, otherwise you get:
```
 UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().
```


It's probably fine to just catch them all and `console.err` them

scripts/compile-igs.js
```diff
@@ -0,0 +1,158 @@
...
+            ignoreVersion = true;
+        } else if (arg === '-h' || arg === 'help') {
+            usage();
+            return;
+        } else {
+            igPath = arg;
+        }
+    }
+    if (igPath === null) {
+        console.error('Need path to IGs folder as argument');
```
carvantes: I think it's ok to have this as an argument for flexibility, but i think that we should make it optional and default to `igPath='../implementationGuides/'` if no argument is provided. IMO most people will use the default location for simplicity

scripts/compile-igs.js
```diff
@@ -0,0 +1,158 @@
...
+        for (const igId of Object.keys(parentMap)) {
+            this.depthFirst([igId], parentMap);
+        }
+    }
+
+    depthFirst(parents, pMap) {
+        const igId = parents[parents.length - 1];
+        const dependencies = pMap[igId];
+        if (!dependencies) {
+            throw new Error(`Missing dependency ${igId}`);
```
carvantes: We need to make an exception for `hl7.fhir.r4.core`. There's no need to compile the base FHIR definitions since they are already already bundled in FWoA.

scripts/compile-igs.js
```diff
@@ -0,0 +1,158 @@
...
+        const index = await loadJson(indexJson);
+        index.files.forEach(file => {
+            if (file.resourceType === 'SearchParameter') {
+                const filePath = path.join(igDir.toString(), file.filename);
+                console.log(`Compiling ${filePath}`);
+                loadJson(filePath).then(jsonData => {
+                    compile(jsonData);
+                });
+            }
+        });
```
carvantes: We need to rearrange this a little bit. `compile` should be called only once with an array of all the search params from all the IGs that must be compiled. My apologies if this wasn't clear before.

Also, the `compile` output must be saved under `./compiledImplementationGuides` as a json file.

carvantes (changes requested):

carvantes:
I updated the PR to merge into `develop` instead of `mainline`. That's the usual procedure for PRs in this repo. Merges from `develop` to `mainline` occur automatically after automated integ tests pass.

The branch switch brought a lot of merge commits into the PR. Looks like this occurred since you branched off from `mainline` instead of from `develop`. It can be cleaned up by doing the following:

1. branch off from development
1. make your changes
1. force push to `INT-931_load_and_parse_IGs` to fix this PR

I'd suggest to clean this up before doing further changes.

Please read our contributing guidelines for more details: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/CONTRIBUTING.md#contributing-via-pull-requests

carvantes:
FYI, I submitted a PR with the implementation of the compile method. https://github.com/awslabs/fhir-works-on-aws-search-es/pull/38

justinusmenzel:
Thank you @carvantes for the great feedback. I just noticed the issue with calling compile for each resource instead of only once with a list of all the resources. 

Question: would it make sense for me to take the original code that I wrote in TypeScript and put it in a ts file and then let the script be basically just a one liner and just call the TS class. That way most of the code is in TS and I can add a test.

carvantes:
> Thank you @carvantes for the great feedback. I just noticed the issue with calling compile for each resource instead of only once with a list of all the resources.
> 
> Question: would it make sense for me to take the original code that I wrote in TypeScript and put it in a ts file and then let the script be basically just a one liner and just call the TS class. That way most of the code is in TS and I can add a test.

That sounds good.

I did something similar on https://github.com/awslabs/fhir-works-on-aws-search-es/pull/38. `scripts/searchParamsCompiler/run.ts` is a small script that imports `SearchImplementationGuides.compile`

I'd suggest to have the small script in typescript as well and run it using [ts-node](https://www.npmjs.com/package/ts-node). It is possible to have it in raw javascript, but that'd require to import typescript-compiled files from `/lib` and to ensure that `tsc` runs before the script and that setup is a bit clunky.

scripts/compile-igs.js
justinusmenzel: Just one JSON file? Should the file be `./compiledImplementationGuides.json` or a JSON file in the `./compiledImplementationGuides` folder. If the latter then what filename should it be? 
`./compiledImplementationGuides/searchParameters.json` maybe, since we are just compiling the `SearchParameter` resources for now. 
Should then different type of resources compiled by different compiles and stored in different files?

justinusmenzel:
I just looked at the `develop` branch and it says there ` This branch is 1 commit ahead, 73 commits behind mainline. ` 
73 commits behind mainline seems like a lot to me. Is that expected?

scripts/compile-igs.js
carvantes: `./compiledImplementationGuides/searchParameters.json` would work. We need a different file per compiler. Actually, something like `./compiledImplementationGuides/fhir-works-on-aws-search-es.json` is best. Naming it after the module that compiled it is better since future compilers may consume several resource types at once.

> Should then different type of resources compiled by different compiles and stored in different files?

yes, that's the plan. But we are starting with only search for now

justinusmenzel: Maybe the ImplementationGuides interface could have a getName() method. That way the compile-igs script just needs to pass a list of ImplementationGuides to the IGCompiler and the compiler can use the getName() method to get the file name like so: `` const fileName = `${igs.getName()}.json` ``

justinusmenzel:
The changes in the MR depend on two pending MRs (https://github.com/awslabs/fhir-works-on-aws-interface/pull/44 and https://github.com/awslabs/fhir-works-on-aws-search-es/pull/38). Do we need to wait for them to be merged before we can update package.json? Right now package.json points to `search-es` and `interface` versions without the changes that this MR depends on.

carvantes:
> The changes in the MR depend on two pending MRs ([awslabs/fhir-works-on-aws-interface#44](https://github.com/awslabs/fhir-works-on-aws-interface/pull/44) and [awslabs/fhir-works-on-aws-search-es#38](https://github.com/awslabs/fhir-works-on-aws-search-es/pull/38)). Do we need to wait for them to be merged before we can update package.json? Right now package.json points to `search-es` and `interface` versions without the changes that this MR depends on.

Yes, those PRs need to be merged and new versions of the NPM packages released before we can update the `package.json` file on this PR. However, you can pull the code from those PRs and test locally

There are more detailed instructions on how to setup the multiple FWoA packages locally for development at: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/DEVELOPMENT.md

scripts/compile-igs.js
carvantes: That's a good idea. Although I'm not making additional changes to the interface for the search-only release. I think that we could further explore this idea once we move to the next phase and start working on additional compilers.

carvantes (changes requested):
I'm running the compile process locally and I have a couple of notes regarding `compile-igs.ts` :

- It should not fail if the `implementationGuides` folder does not exist. If it does not exist the script should simply be a no op since there's nothing to compile. Maybe just print a console message saying that no IGs were found.
- It should not fail if the `compiledImplementationGuides` folder does not exist. If it does not exist, the script should create it as part of writing the output file.

src/IGCompiler.test.ts
```diff
@@ -0,0 +1,179 @@
...
+        });
+    }
+}
+
+interface IGVersion {
+    version: string;
+    deps: string[];
+}
+
+describe('IGCompiler tests', function() {
```
carvantes: Could you fix the eslint warnings? They don't fail the build but so far this package doesn't have any warnings and it'd be nice to keep it that way. For all these warnings just use arrow functions instead `() => {}`

src/IGCompiler.ts
```diff
@@ -0,0 +1,168 @@
...
+     * @param igsDir
+     * @param outputPath
+     */
+    public async compileIGs(igsDir: PathLike, outputPath: PathLike): Promise<void> {
+        const resources: any[] = [];
+        if (!existsSync(igsDir)) {
+            throw new Error(`'${igsDir}' doesn't exist`);
+        }
+        for (const igInfo of await this.collectIGInfos(igsDir)) {
+            await this.collectResources(igInfo.path, resources);
```
carvantes: Can we change `collectResources` to return an array and omit the `resources` parameter? I think that creating an empty array on L90 shouldn't be necessary

src/IGCompiler.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    constructor(implementationGuides: ImplementationGuides, options: IGCompilerOptions) {
+        this.options = options;
+        this.implementationGuides = implementationGuides;
+    }
+
+    async collectResources(igDir: PathLike, resources: any[]): Promise<void> {
+        const indexJson = path.join(igDir.toString(), '.index.json');
+        if (!existsSync(indexJson)) {
+            return;
```
carvantes: I think we should throw an error here. 

This condition will rarely happen, but it would mean that the IG cannot be processed. It'd be better to fail the compilation instead of skipping an entire IG silently.

src/IGCompiler.ts
```diff
@@ -0,0 +1,168 @@
...
+        if (this.options.ignoreVersion) {
+            return name;
+        }
+        return `${name}@${version}`;
+    }
+
+    async extractIgInfo(igDir: PathLike): Promise<IGInfo | null> {
+        const packagePath = path.join(igDir.toString(), 'package.json');
+        if (!existsSync(packagePath)) {
+            return null;
```
carvantes: We should throw an error here. Same reason as my other comment regarding a missing `index.json` file

src/IGCompiler.ts
```diff
@@ -0,0 +1,168 @@
...
+            if (igInfo) {
+                if (igInfo.name !== BASE_FHIR_NAME) {
+                    igInfos.push(igInfo);
+                }
+                parentMap[igInfo.id] = igInfo.dependencies;
+            }
+        }
+        for (const igId of Object.keys(parentMap)) {
+            this.depthFirst([igId], parentMap);
+        }
```
carvantes: can we refactor this into a separate private method?

I'm thinking of `validateDependencies(igs: IGInfo[]): void`. That method builds the `parentMap` and calls `depthFirst`

I think that makes it clearer that the intention is to validate. Also, I was a bit confused about `parentMap` at first but now I see that it is an auxiliary data structure only meant to be used alongside `depthFirst`. Having those 2 things inside of `validateDependencies` would make it clearer.

src/IGCompiler.ts
```diff
@@ -0,0 +1,168 @@
...
+    async collectIGInfos(igsDir: PathLike): Promise<IGInfo[]> {
+        const igInfos: IGInfo[] = [];
+        const parentMap: { [key: string]: string[] } = {};
+        for (const igPath of await listIgDirs(igsDir)) {
+            console.log(`looking at ig path: ${igPath}`);
+            const igInfo = await this.extractIgInfo(igPath);
+            if (igInfo) {
+                if (igInfo.name !== BASE_FHIR_NAME) {
+                    igInfos.push(igInfo);
+                }
```
carvantes: I think we should add an `else` here and log a message like `Skipping hl7.fhir.r4.core since the base FHIR definitions are already included in fhir-works-on-aws`. It may be confusing for some users if we just skip it silently.

Also, I'm getting an error when trying to compile us core: `Exception:  Missing dependency hl7.fhir.r4.core@4.0.1 Error: Missing dependency hl7.fhir.r4.core@4.0.1`. This shouldn't happen since `hl7.fhir.r4.core` is already included in FWoA. We should ignore it on the dependency check.

Apologies if my previous comment about making an exception for `hl7.fhir.r4.core` was not clear about the dependencies part.

carvantes (changes requested):

src/IGCompiler.ts
justinusmenzel: I can change the code to:
```
        const resources: any[] = [];
        for (const igInfo of await this.collectIGInfos(igsDir)) {
            resources.push(...(await this.collectResources(igInfo.path)));
        }
```
I don't see how I could get rid of creating the empty array in L90.

src/IGCompiler.ts
justinusmenzel: For testing it was convenient to point the compiler to a node_modules folder where some IGs were installed next to some other dependencies.

src/IGCompiler.ts
carvantes: Your proposed change looks good. 

you are right, an array is needed to collect all the resources. My bad, didn't notice that `collectResources` was being called within a loop.

carvantes:
@justinusmenzel, I just merged this PR into a feature branch. There are still a few more things that need to be done for this feature before it can be merged back into `develop`

Thanks again for contributing!

PR #205, opened 2021-02-10
feat: add Implementation Guides integ tests

carvantes:
Description of changes:

GH actions will download US Core and Compile it so that it is loaded by FWoA. Integ tests then will validate that US core specific search params are working. But right now, for search parameters alone, it is fine

There were some discussions earlier during the design review about whether or not we nee a separate environment for IGs integ tests. For search tests alone it's totally fine to use the same environment. We may revisit that if/when the IGs implementation brings up different pieces of infrastructure.

Checklist:

### DIFF
--- a/integration-tests/implementationGuides.test.ts
+++ b/integration-tests/implementationGuides.test.ts
@@ -19,9 +19,11 @@ describe('Implementation Guides - US Core', () => {
         const capabilityStatement: CapabilityStatement = (await client.get('metadata')).data;
 
         const usCorePatientSearchParams = capabilityStatement.rest[0].resource
-            .filter(x => x.type === 'Patient')
-            .flatMap(x => x.searchParam ?? [])
-            .filter(x => x.definition.startsWith('http://hl7.org/fhir/us/core/SearchParameter/us-core'));
+            .filter(resource => resource.type === 'Patient')
+            .flatMap(resource => resource.searchParam ?? [])
+            .filter(searchParam =>
+                searchParam.definition.startsWith('http://hl7.org/fhir/us/core/SearchParameter/us-core'),
+            );
 
         expect(usCorePatientSearchParams).toEqual(
             // There are many more search parameters in US Core but they are all loaded into FWoA in the same way.
@@ -58,8 +60,6 @@ describe('Implementation Guides - US Core', () => {
                 },
             ]),
         );
-
-        // console.log(usCorePatientSearchParams);
     });
 
     test('query using search parameters', async () => {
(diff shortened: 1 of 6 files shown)


integration-tests/implementationGuides.test.ts
```diff
@@ -0,0 +1,136 @@
...
+describe('Implementation Guides - US Core', () => {
+    let client: AxiosInstance;
+    beforeAll(async () => {
+        client = await getFhirClient();
+    });
+    test('capability statement includes search parameters', async () => {
+        const capabilityStatement: CapabilityStatement = (await client.get('metadata')).data;
+
+        const usCorePatientSearchParams = capabilityStatement.rest[0].resource
+            .filter(x => x.type === 'Patient')
```
nguyen102: Nitpick: To make it more readable we can replace `x` with `resource`

integration-tests/implementationGuides.test.ts
```diff
@@ -0,0 +1,136 @@
...
+                        ' - comparator<br />\n' +
+                        ' - modifier<br />\n' +
+                        ' - chain<br />\n' +
+                        '\n' +
+                        ' ',
+                },
+            ]),
+        );
+
+        // console.log(usCorePatientSearchParams);
```
nguyen102: delete this line?

integration-tests/types.ts
```diff
@@ -0,0 +1,125 @@
...
+export interface REST {
+    mode: string;
+    documentation: string;
+    security: Security;
+    resource: Resource[];
+    interaction: Interaction[];
+    operation: Operation[];
+}
+
+export interface CapabilityStatement {
```
nguyen102, line 110: Should we add this to the interface package? Looks like we're also using this type when generating the cap statement.
https://github.com/awslabs/fhir-works-on-aws-routing/blob/5a61db3ac3b50116bdd119b98a929065676a0d0a/src/router/metadata/cap.template.ts

integration-tests/implementationGuides.test.ts
carvantes: I agree. Done.

integration-tests/implementationGuides.test.ts
carvantes: oops, missed that. Done

integration-tests/types.ts
carvantes: I don't think so. 

Integ tests should not share code/types with the internal implementation. These are just auto generated types from a sample JSON cap statement. They are good enough for tests.

nguyen102 (approved):

PR #214, opened 2021-02-16
fix: Update oauth2ApiEndpoint

nguyen102:
Issue #, if available:

Description of changes:
The AuthZ server that was set up in Okta and was also deployed by Patient Picket has the id `aus4az2kcEfXCbGmI5d6`, not `default`. 

This change is needed so that we get the proper `jwks` at https://dev-80793861.okta.com/oauth2/aus4az2kcEfXCbGmI5d6/v1/keys not https://dev-80793861.okta.com/oauth2/default/v1/keys.

This will allow us to make authenticated requests to the FHIR Server.


nguyen102:
issuerEndpoint and oauth2ApiEndpoint are now provided through env variables. I'm closing this PR and updating the GH secrets instead.

PR #221, opened 2021-02-23
feat: add HAPI validator lambda fn

carvantes:
Description of changes:

Add the java lambda function that runs the HAPI FHIR Validator.

It is expected that the main compiler script copies the IGs to `javaHapiValidatorLambda/src/main/resources/implementationGuides` so that they are bundled into the lambda deployment package an loaded to the HAPI validator.

Checklist:


javaHapiValidatorLambda/serverless.yml
```diff
@@ -0,0 +1,21 @@
...
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+service: fhir-service-validator
+
+provider:
+  name: aws
+  region: us-west-2
+  stage: dev
+  runtime: java8
```
nguyen102: Any reason why not java11? Is the validator built for java8?
https://aws.amazon.com/about-aws/whats-new/2019/11/aws-lambda-supports-java-11/

javaHapiValidatorLambda/serverless.yml
```diff
@@ -0,0 +1,21 @@
...
+
+package:
+  artifact: target/fwoa-hapi-validator-dev.jar
+
+functions:
+  hello:
```
nguyen102: Should we change the name of this function? Maybe something like `validate`?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,159 @@
...
+        PrePopulatedValidationSupport prepopulatedValidationSupport = loadIgs(ctx);
+        supportChain.addValidationSupport(prepopulatedValidationSupport);
+
+        // Wrap the chain in a cache to improve performance
+        CachingValidationSupport cache = new CachingValidationSupport(supportChain);
+
+        // Create a validator using the FhirInstanceValidator module.
+        FhirInstanceValidator validatorModule = new FhirInstanceValidator(cache);
+        validator = ctx.newValidator().registerValidatorModule(validatorModule);
+
```
nguyen102: nitpick: Delete extra line

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    public ValidatorResponse validate(String resourceAsJsonText) {
+        try {
+            ValidationResult result = validator.validateWithResult(resourceAsJsonText);
+            return toValidatorResponse(result);
+
```
nguyen102: nitpick: Delete extra line

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,159 @@
...
+                                    break;
+                                default:
+                                    // cannot happen since we checked for allowedResourceTypes
+                                    break;
+                            }
+                        }
+                    }
+                }
+            }
+        } catch (IOException e) {
```
nguyen102: If we can't read one of the files, should we error out here, and stop execution or should we only report an error happened?

What happens if the validator expects to load in a file, but fails to do it. Will the validator not validate against that IG?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/ValidatorResponse.java
```diff
@@ -0,0 +1,25 @@
...
+    private boolean isSuccessful;
+    private List<ValidatorErrorMessage> errorMessages;
+}
+
+@Builder
+@Value
+class ValidatorErrorMessage {
+    private String severity;
+    private String msg;
+}
```
nguyen102: Do we need an extra line here?

javaHapiValidatorLambda/src/test/java/software/amazon/fwoa/ValidatorTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.google.common.collect.ImmutableList;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ValidatorTest {
```
nguyen102, line 16: Do we need a test to validate that we can correctly load an IG and validate a resource with that IG?

nguyen102 (commented):
Overall looks good. I have some formatting suggestions and a question about the try/catch.

Btw, what are your thoughts on moving this code out into it's own package? I know we initially designed this to be similar to the `auditLogMover`. However, it's a big mental shift to change from looking at TS/JS code in this package to Java. This code has very clear separations of concern from the rest of the `deployment` package and it seems like it could exist as it's own package.

For now, it makes sense if we're just including the code in this package to make development easier. I'm ok with keeping it here for now, with the plan to move it to its own package in the future.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/ValidatorResponse.java
carvantes: added new line

javaHapiValidatorLambda/serverless.yml
carvantes: renamed to validate. hello was a remnant from the base java app generated by serverless

javaHapiValidatorLambda/serverless.yml
carvantes: I changed it to java11

Looks like HAPI targets java 8: https://github.com/hapifhir/hapi-fhir/blob/master/pom.xml#L1863-L1864, so it's not really taking advantage of new features of java 11. But at the same time, building && running on java 11 is not causing any trouble so far.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: done

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: done

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: I changed it to just rethrow the exception. 

Since we are using provisioned concurrency, failing to load an IG will be an uncaught exception that will make the deployment fail. This is an ok behavior since that means that the IG files are wrong and you need to fix them and redeploy.

javaHapiValidatorLambda/src/test/java/software/amazon/fwoa/ValidatorTest.java
carvantes: That will be covered by integ tests.

The thing is that the IGs are added to `src/main/resources/implementationGuides` as if they were part of the lambda source code and it is very tricky to mock those files in a way that makes sense.

carvantes:
> Btw, what are your thoughts on moving this code out into it's own package? I know we initially designed this to be similar to the `auditLogMover`. However, it's a big mental shift to change from looking at TS/JS code in this package to Java. This code has very clear separations of concern from the rest of the `deployment` package and it seems like it could exist as it's own package.
> 
> For now, it makes sense if we're just including the code in this package to make development easier. I'm ok with keeping it here for now, with the plan to move it to its own package in the future.

yeah, we can think about it. I'm not quite sure what's best approach. There's no straightforward way to "import" a package that has CFN templates + code (and java code only makes it harder).

IMO it's ok to keep it here in the short term.

javaHapiValidatorLambda/serverless.yml
```diff
@@ -0,0 +1,35 @@
...
+functions:
+  validator:
+    handler: software.amazon.fwoa.Handler
+    timeout: 60
+    memorySize: 2048
+    provisionedConcurrency: 5
+
+resources:
+  - Parameters:
+      Stage:
```
rsmayda, line 35: Does this `Stage` actually work and change the stage of the stack? i.e. this is how we do it in the main serverless: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L17

javaHapiValidatorLambda/serverless.yml
```diff
@@ -0,0 +1,35 @@
+#
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+service: fhir-service-validator
+
+provider:
```
rsmayda, line 12: Could you add logRetention and StackTag? https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L28-L30

javaHapiValidatorLambda/serverless.yml
```diff
@@ -0,0 +1,35 @@
...
+  name: aws
+  region: us-west-2
+  stage: dev
+  runtime: java11
```
rsmayda, line 16: Could you add x-ray tracing too: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L56

Will be helpful when we do a load test

javaHapiValidatorLambda/pom.xml
```diff
@@ -0,0 +1,185 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
rsmayda, line 1: Did we check the licenses of these new dependencies?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,161 @@
...
+@Slf4j
+public class Validator {
+    private static final Gson GSON = new Gson();
+    public static final String IMPLEMENTATION_GUIDES_FOLDER = "implementationGuides";
+
+    private final FhirValidator validator;
+
+    public Validator() {
+        // To learn more about the different ways to configure FhirInstanceValidator see: https://hapifhir.io/hapi-fhir/docs/validation/validation_support_modules.html
+        FhirContext ctx = FhirContext.forR4();
```
rsmayda, line 53: Pass in version to have R4 or STU3?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,161 @@
...
+import software.amazon.fwoa.models.IgIndex;
+
+/**
+ * This class is a wrapper around the HAPI FhirValidator.
+ * The FhirValidator is built using default settings and the available implementation guides are loaded into it.
+ */
+@Slf4j
+public class Validator {
+    private static final Gson GSON = new Gson();
+    public static final String IMPLEMENTATION_GUIDES_FOLDER = "implementationGuides";
```
rsmayda, line 47: Could/should this be passed in?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -0,0 +1,161 @@
...
+        final PrePopulatedValidationSupport prePopulatedValidationSupport = new PrePopulatedValidationSupport(ctx);
+
+        try (ScanResult allFiles = new ClassGraph().acceptPaths(IMPLEMENTATION_GUIDES_FOLDER).rejectPaths(IMPLEMENTATION_GUIDES_FOLDER + "/*/*").scan()) {
+            ResourceList jsonResources = allFiles.getResourcesWithExtension("json");
+
+            ResourceList indexFiles = jsonResources.filter(x -> x.getPath().endsWith(".index.json"));
+
+            for (Resource indexFile : indexFiles) {
+                IgIndex igIndex = GSON.fromJson(indexFile.getContentAsString(), IgIndex.class);
+                for (IgFile file : igIndex.files) {
```
rsmayda, line 122: Value in parallelization of these? How many files would we loop through here. The processing looks independent to each other

rsmayda (changes requested):

javaHapiValidatorLambda/serverless.yml
carvantes: done

javaHapiValidatorLambda/serverless.yml
carvantes: done

javaHapiValidatorLambda/serverless.yml
carvantes: Not really. I changed it to use the `custom:` serverless params so that it matches the deploy experience of the deployment package.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: It could be, but there's no use case right now for looking for IGs at a different location.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: It's not safe to do so.
Ultimately we want to load all files into `PrePopulatedValidationSupport`, but that class is not advertised as being thread safe. looking at it's source code, It relies on several `HashMap`(instead of something like `ConcurrentHashMap `) so it is indeed not thread safe.

javaHapiValidatorLambda/pom.xml
carvantes: Yes, they are ok.

They look like many packages, but most come from either aws-lambda-java-libs or HAPI that are both Apache 2.0

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: I'll add support for STU3 later. 

Right now there is a small bug that prevents loading ValueSets from IGs into the validator
https://chat.fhir.org/#narrow/stream/179167-hapi/topic/PrePopulatedValidationSupport/near/217423371

rsmayda (approved):

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
carvantes: https://github.com/hapifhir/hapi-fhir/issues/2432

PR #222, opened 2021-02-23
chore: change log retention and add DDB tables as output

rsmayda:
Log retention being bump to 5 years as 90 days was too short and by default we should be providing a safe choice

Checklist:

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -524,7 +524,7 @@ resources:
         Description: DynamoDB table for storing non-Binary resources
         Value: !GetAtt ResourceDynamoDBTableV2.Arn
       ResourceDynamoDBTableV2StreamArn:
-        Description: DynamoDB stream for storing non-Binary resources catching all updates to the DDB
+        Description: DynamoDB stream for the DDB table storing non-Binary resources
         Value: !GetAtt ResourceDynamoDBTableV2.StreamArn
       ExportRequestDynamoDBTableArn:
         Description: DynamoDB table for storing bulk export requests
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -522,10 +522,13 @@ resources:
         Value: !Ref FHIRBinaryBucket
       ResourceDynamoDBTableV2Arn:
         Description: DynamoDB table for storing non-Binary resources
-        Value: !GetAtt ResourceDynamoDBTableV2.Arn 
+        Value: !GetAtt ResourceDynamoDBTableV2.Arn
+      ResourceDynamoDBTableV2StreamArn:
+        Description: DynamoDB stream for storing non-Binary resources catching all updates to the DDB
+        Value: !GetAtt ResourceDynamoDBTableV2.StreamArn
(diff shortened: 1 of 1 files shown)


serverless.yaml
```diff
@@ -520,6 +520,12 @@ resources:
       FHIRBinaryBucket:
         Description: S3 Bucket for storing Binary Objects
         Value: !Ref FHIRBinaryBucket
+      ResourceDynamoDBTableV2Arn:
+        Description: DynamoDB table for storing non-Binary resources
+        Value: !GetAtt ResourceDynamoDBTableV2.Arn
```
carvantes: From the email, it sounds like they want us to export the `StreamArn` as well

carvantes (dismissed):

serverless.yaml
rsmayda: Yep I think you are right

serverless.yaml
```diff
@@ -520,6 +520,15 @@ resources:
       FHIRBinaryBucket:
         Description: S3 Bucket for storing Binary Objects
         Value: !Ref FHIRBinaryBucket
+      ResourceDynamoDBTableV2Arn:
+        Description: DynamoDB table for storing non-Binary resources
+        Value: !GetAtt ResourceDynamoDBTableV2.Arn
+      ResourceDynamoDBTableV2StreamArn:
+        Description: DynamoDB stream for storing non-Binary resources catching all updates to the DDB
```
rsmayda:
```suggestion
        Description: DynamoDB stream for the DDB table storing non-Binary resources
```

carvantes (approved):

R-K-1:
Hello,

I see the resource table and stream ARNs in output of [serverless.yaml](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml). I have been deploying the stack by directly using the CFN template via the following command

```
cloudformation.create_stack(
    StackName=fhir_stack_name,
    TemplateURL='https://solutions-reference.s3.amazonaws.com/fhir-works-on-aws/latest/fhir-works-on-aws.template',
    TimeoutInMinutes=123,
    Capabilities=[
        'CAPABILITY_IAM','CAPABILITY_NAMED_IAM','CAPABILITY_AUTO_EXPAND',
    ],
    OnFailure='ROLLBACK',
)
```
However, the output of the stack generated by the command above does not contain the ARNs mentioned earlier? Is this CFN template not linked to serverless.yaml?

rsmayda:
Hey, thanks for reaching out! The template you are referencing comes from AWS Solutions which is on a slower slower release cadence, but that code is referenced under the `aws-solution` branch: https://github.com/awslabs/fhir-works-on-aws-deployment/tree/aws-solution

If you would like to get this fix ASAP we recommend you installing from [mainline](https://github.com/awslabs/fhir-works-on-aws-deployment). For steps in the install process check out the [INSTALL README](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/INSTALL.md)

PR #224, opened 2021-02-24
test: Add SMART negative tests

nguyen102:
Issue #, if available:

Description of changes:
* Add SMART Negative tests
* Also migrated search tests to run with SMART
* Bulk Export tests are current skipped until this [bug](https://github.com/awslabs/fhir-works-on-aws-authz-smart/issues/28) is fixed

**Testing**
Tests passed when running against SMART test environment.

### DIFF
--- a/integration-tests/smartAuth.test.ts
+++ b/integration-tests/smartAuth.test.ts
@@ -4,9 +4,8 @@ import { getFhirClient, idsOfFhirResources } from './utils';
 const sherlockId = idsOfFhirResources.sherlockHolmes;
 const mycroftId = idsOfFhirResources.mycroftHolmes;
 
-async function getPatient(client: AxiosInstance, patientId: string | undefined = undefined) {
-    const url = patientId ? `/Patient/${patientId}` : '/Patient';
-
+async function getPatient(client: AxiosInstance, patientId: string) {
+    const url = `/Patient/${patientId}`;
     return client.get(url);
 }
 
--- a/integration-tests/smartAuth.test.ts
+++ b/integration-tests/smartAuth.test.ts
@@ -1,5 +1,5 @@
-import axios from 'axios';
-import { stringify } from 'query-string';
+import { AxiosInstance } from 'axios';
+import { getFhirClient } from './utils';
 
 /*
  *  DDB Set Up on SMART Integration Env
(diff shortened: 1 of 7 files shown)


integration-tests/smartAuth.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        const fhirClient = await getFhirClient('launch/patient patient/Patient.read', false, 'Invalid Access Token');
+        await expect(getPatient(fhirClient, sherlockId)).rejects.toThrowError(
+            new Error('Request failed with status code 403'),
+        );
+    });
+
+    test("Failed to read Watson's record, because of insufficient scope", async () => {
+        const watsonId = '7965ea12-7ecd-46cd-9ec1-340400c9548c';
+        // FhirClient does not include enough permission because patient/Patient.read scope is missing
+        const fhirClient = await getFhirClient('launch/patient patient/Patient.read', false);
```
carvantes: Is the comment correct? It looks like `patient/Patient.read` is part of the requested scopes.
Why does this test fails?

integration-tests/smartAuth.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        await expect(getPatient(fhirClient, sherlockId)).rejects.toThrowError(
+            new Error('Request failed with status code 403'),
+        );
+    });
+
+    test('Invalid access token', async () => {
+        const fhirClient = await getFhirClient('launch/patient patient/Patient.read', false, 'Invalid Access Token');
+        await expect(getPatient(fhirClient, sherlockId)).rejects.toThrowError(
+            new Error('Request failed with status code 403'),
```
carvantes: Make assertions on the numeric status code instead of the string error message

i.e.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/6ff3899563f66db6193ac9cbb33c2059a78eaba4/integration-tests/search.test.ts#L84-L86

integration-tests/smartAuth.test.ts
```diff
@@ -0,0 +1,66 @@
...
+const sherlockId = '92e0d921-bb19-4cae-a3cc-9d3c5bcf7a39';
+const mycroftId = 'cf52937f-a28f-437e-bfac-2228f5db6801';
+
+async function getPatient(client: AxiosInstance, patientId: string | undefined = undefined) {
+    const url = patientId ? `/Patient/${patientId}` : '/Patient';
+
+    return client.get(url);
+}
+
+test('Successfully read Sherlock and Mycroft records', async () => {
```
carvantes: I think that for the test names it's better to avoid names like Sherlock and use their "roles" instead. i.e. "Successfully read own user record and records that reference the user"

integration-tests/smartAuth.test.ts
nguyen102: Ah, the test is right, but the comment is incorrect. Since Watson doesn't have any references to Sherlock, Sherlock cannot read Watson's record.

integration-tests/smartAuth.test.ts
nguyen102: updated

integration-tests/smartAuth.test.ts
nguyen102: updated

integration-tests/smartAuth.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import { getFhirClient, idsOfFhirResources } from './utils';
+
+const sherlockId = idsOfFhirResources.sherlockHolmes;
+const mycroftId = idsOfFhirResources.mycroftHolmes;
+
+async function getPatient(client: AxiosInstance, patientId: string | undefined = undefined) {
+    const url = patientId ? `/Patient/${patientId}` : '/Patient';
+
+    return client.get(url);
+}
```
carvantes: I don't like that this `getPatient` can be a READ or a SEARCH depending on the presence of `patientId`

looks like it is never used without a patientId parameter. Plus, if you wanted to search with this fn there's no way to pass search params. 

I'd rather remove the `patientId` param or maybe even remove this function altogether(It's fairly simple to just call `fhirClient.get('Patient/${sherlockId}');`)

carvantes (dismissed):

integration-tests/smartAuth.test.ts
nguyen102: You're right. This function can be updated to remove the case in which no `patientId` is passed to it. This was left over for when I was using this function to also do searches. I still prefer to keep this function though, as I find it makes the code slightly more readable to have that function then to know to call `/Patient/${patientId}`

carvantes (approved):

PR #229, opened 2021-03-03
test: Add tests for RBAC Auth

nguyen102:
Issue #, if available:

Description of changes:

Added integration tests for RBAC Auth

Checklist:

### DIFF
--- a/integration-tests/bulkExport.test.ts
+++ b/integration-tests/bulkExport.test.ts
@@ -6,6 +6,7 @@ import BulkExportTestHelper, { ExportStatusOutput } from './bulkExportTestHelper
 import { getFhirClient } from './utils';
 
 const FIVE_MINUTES_IN_MS = 5 * 60 * 1000;
+jest.setTimeout(FIVE_MINUTES_IN_MS);
 
 describe('Bulk Export', () => {
     let bulkExportTestHelper: BulkExportTestHelper;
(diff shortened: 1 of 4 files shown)


integration-tests/utils.ts
```diff
@@ -7,8 +7,19 @@ import * as AWS from 'aws-sdk';
 import axios from 'axios';
 import { Chance } from 'chance';
 
-export const getFhirClient = async () => {
-    const { API_URL, API_KEY, API_AWS_REGION, COGNITO_USERNAME, COGNITO_PASSWORD, COGNITO_CLIENT_ID } = process.env;
+export const getFhirClient = async (
+    isAuditor: boolean = false,
+    providedAccessToken: string | undefined = undefined,
```
carvantes: Can we use strings for the role instead? Otherwise it's hard to understand the meaning of something like `getFhirClient(false)`
Also, `providedAccessToken` is effectively an optional parameter
```suggestion
    role: 'auditor'|'practitioner' = 'auditor',
    providedAccessToken: string?
```

integration-tests/bulkExport.test.ts
```diff
@@ -6,6 +6,7 @@ import BulkExportTestHelper, { ExportStatusOutput } from './bulkExportTestHelper
 import { getFhirClient } from './utils';
```
rsmayda, line 6: just reformatting here right?

nguyen102: It's mostly reformatting. I'm setting the jest timeout for all tests in this file on line 9, instead of passing that value to each test function. As a result the indentation of the lines are changed.

integration-tests/utils.ts
nguyen102: updated

integration-tests/rbac-permission.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    delete patientRecord.id;
+    await expect(client.post('Patient', patientRecord)).resolves.toMatchObject({
+        status: 201,
+        data: patientRecord,
+    });
+});
+
+describe('Negative tests', () => {
+    test('invalid token', async () => {
+        const client = await getFhirClient('practitioner', 'Invalid token');
```
carvantes, line 17: Can we add more tests with invalid tokens? `Invalid token` gets rejected right away since it's syntactically invalid. I think we should add a couple more tests with syntactically valid tokens like:

- Invalid signature:`<valid header>.<valid claims. set exp to a date in the far distant future>.xxxInvalidSignaturexxx`
- Attempt to use an unsafe algorithm: `<valid header with alg=none>.<valid claims>` 

Actually these tests far more relevant for SMART since we own the token validation step. It's fair to assume that Cognito just works.

carvantes (approved):

rsmayda (approved):

PR #237, opened 2021-03-08
test: IGs: Validation during create/update and supportedProfile in CapStatement

nguyen102:
Issue #, if available:

Description of changes:

test: IGs: Validation during create/update and supportedProfile in CapStatement

Checklist:

### DIFF
--- a/integration-tests/implementationGuides.test.ts
+++ b/integration-tests/implementationGuides.test.ts
@@ -15,16 +15,33 @@ describe('Implementation Guides - US Core', () => {
     beforeAll(async () => {
         client = await getFhirClient();
     });
-    test('capability statement includes search parameters', async () => {
-        const capabilityStatement: CapabilityStatement = (await client.get('metadata')).data;
 
-        const usCorePatientSearchParams = capabilityStatement.rest[0].resource
+    function getResourcesWithSupportedProfile(capStatement: CapabilityStatement) {
+        const resourcesWithSupportedProfile: Record<string, string[]> = {};
+        capStatement.rest[0].resource
+            .filter(resource => {
+                return resource.supportedProfile;
+            })
+            .forEach(resource => {
+                if (resource.type) {
+                    resourcesWithSupportedProfile[resource.type] = resource.supportedProfile!.sort();
+                }
+            });
+
+        return resourcesWithSupportedProfile;
+    }
+
+    test('capability statement includes search parameters and supportedProfile', async () => {
+        const actualCapabilityStatement: CapabilityStatement = (await client.get('metadata')).data;
+
+        const usCorePatientSearchParams = actualCapabilityStatement.rest[0].resource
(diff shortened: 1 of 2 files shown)


integration-tests/implementationGuides.test.ts
```diff
@@ -93,14 +128,97 @@ describe('Implementation Guides - US Core', () => {
...
+        await expect(client.post('Patient', patient)).rejects.toMatchObject({
+            response: noTextFieldErrorResponse,
+        });
+    });
+
+    test('query using search parameters', async () => {
+        const patient = getRandomPatientWithEthnicityAndRace();
 
-        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', usCorePatient)).data;
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', patient)).data;
```
carvantes, line 222: I think that these change will cause the search tests to eventually fail.

Tests relied on the fact that the `raceCode` was unique to ensure that the patient was part of search results. Now that the test data is always using the same `raceCode`, the query will return several results and as Patients accumulate the expected Patient will get pushed out of the first search results page and fail the test.

btw this will be a lot easier to fix if we actually supported searching by `_lastUpdated` or using `_sort` to get the newest resources first 😉

nguyen102: The test was changed to use the same `raceCode` because creating an invalid `raceCode` causes validation to fail during the create step. 

```
{
    "resourceType": "OperationOutcome",
    "text": {
        "status": "generated",
        "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><h1>Operation Outcome</h1><table border=\"0\"><tr><td style=\"font-weight: bold;\">error</td><td>[]</td><td><pre>Patient.extension[0].extension[0].value.ofType(Coding) - Unknown code 'urn:oid:2.16.840.1.113883.6.238#abc' for 'urn:oid:2.16.840.1.113883.6.238#abc'</pre></td></tr></table></div>"
    },
    "issue": [
        {
            "severity": "error",
            "code": "invalid",
            "diagnostics": "Patient.extension[0].extension[0].value.ofType(Coding) - Unknown code 'urn:oid:2.16.840.1.113883.6.238#abc' for 'urn:oid:2.16.840.1.113883.6.238#abc'"
        }
    ]
}
```

What are your thoughts on narrowing the search result to one result by adding name as part of the search parameter? For example: 
```
p({ race: raceCode, name: testPatient.name[0].family }),
p({ ethnicity: ethnicityCode, name: testPatient.name[0].family }),
```

This isn't ideal because we're searching on more than one parameter. However, it does still confirm that searching on `race` and `ethnicityCode` works.

carvantes: that works

carvantes (approved):

Issue #239, opened 2021-03-09
RE: Security Errors and Deprecation Warnings in latest branches


segfahlt:
Hey guys,

Looking to get this up and running at my company.  When I clone the latest and checkout any of these branches [Mainline, Aws Solution, Develop] and run npm install, I get a ton of deprecation warnings and severe security issues warnings.  There is no way my compliance department will buy off on that.

Any advice or fixes?

Thanks!

Here is a paste of some of the warnings/errors:

PS C:\Development\FHIR\aws\fhir-works-on-aws-deployment> npm install
npm WARN deprecated @hapi/boom@7.4.11: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/joi@15.1.1: Switch to 'npm install joi'
npm WARN deprecated @hapi/hapi@18.4.1: This version contains severe security issues and defects and should not be used! Please upgrade to the latest version of @hapi/hapi or consider a commercial license (https://github.com/hapijs/hapi/issues/4114)
npm WARN deprecated @hapi/hoek@8.5.1: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/joi@16.1.8: Switch to 'npm install joi'
npm WARN deprecated @hapi/wreck@15.1.0: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/bourne@1.3.2: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/topo@3.1.6: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/address@2.1.4: Moved to 'npm install @sideway/address'
npm WARN deprecated @hapi/ammo@3.1.2: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/bounce@1.3.2: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/call@5.1.3: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/catbox-memory@4.1.1: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/heavy@6.2.2: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/catbox@10.2.3: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/mimos@4.1.1: This version has been deprecated and is no longer supported or maintained
[...]

Issue #234, opened 2021-03-05
Error during win_install.ps1 execution - Lambda function is too large


rye442:
I am trying to use FHIRWorks with Windows Install. when I run scripts/win_install.ps1, I get the InvalidParameterValueException.

I checked #184 and #166 before writing this issue. ```npm prune``` and ```npm dedupe``` were also tried, but they did not solve the problem.

I was not able to try the following commands because there was no awk command in my environment.
```sudo npm list -g --depth=0. | awk -F ' ' '{print $2}' | awk -F '@' '{print $1}' | sudo xargs npm remove -g```

## Run commands

```
npm prune
npm dedupe
.\scripts\win_install.ps1 -Region us-east-2 -Stage dev
```

## Error ditails
```
  Serverless Error ---------------------------------------

  An error occurred: UpdateStatusLambdaFunction - Unzipped size must be smaller than 262144000 bytes (Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException; Request ID: eb10a477-29a7-4162-bab4-3d9a498564b9; Proxy: null).
```

## Environment Information
```
  Your Environment Information ---------------------------
     Operating System:          win32
     Node Version:              14.4.0
     Framework Version:         1.83.2 (standalone)
     Plugin Version:            3.8.4
     SDK Version:               2.3.2
     Components Version:        2.34.9
```

## Note
I got error #233 when I ran win_install.ps1, but this was not an error that could not be continued. For this reason, I ignored the #233 error and continue the program, which confirmed this error. I don't think this is relevant, but just in case, I'll mention it.

Bingjiling:
Hi @rye442 , thanks for sharing the details. I'll try recreating the issue on my end for Windows deployment. At the meantime, you could use [Docker installation](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/INSTALL.md#docker-installation) as a work around.

rye442:
Hi @Bingjiling, thanks for the reply. I am going to try Docker installation.

However, keep in mind that even if I can Docker installation, it will not solve the problem of not being able to Windows installation.

Bingjiling:
Hi @rye442, thanks for the update. I recreated the issue with a Windows environment. We'll look further into the issue and keep you posted on the progress.

nguyen102:
Hi @rye442 I currently have code in PR that might fix your issue.

[...]

PR #240, opened 2021-03-10
chore: Update serverless dependencies and use serverless-bundle to optimize package size

nguyen102:
Issue #, if available:

Description of changes:
* Update `serverless` to version `2.29.0` to resolve security warnings. [Related issue](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/239)
* Added `serverless-bundle` to reduce package size. On MacOs this **reduce package size from 18 Mb to 3.5MB**. On windows this reduce package size from 75MB to 3.5 MB. This fix will also resolve the issue [here](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/234)
* Updated Windows install script
   * We need to specify the node version that is installed since our package doesn't support the latest node package which is version 15. That resolve this issue `error concat-text-webpack-plugin@0.1.9: The engine "node" is incompatible with this module. Expected version ">=8 <=14". Got "15.11.0`
   * We need to install `serverless` with `yarn` instead of Windows package manager `choco`. This resolve the webpack issue shown [here](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/406#issuecomment-687200253)

*Testing*
* Window install was successful on a Window Server using the Window's install script
* FHIR deploy from MacOS still passes
* Local development using `yalc` script still works

Checklist:


auditLogMover/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "target": "es2019", // Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'.
-    "module": "commonjs", // Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'.
     "strict": true, // // Enable all strict type-checking options.
+    "moduleResolution": "node",
```
nguyen102, line 5: Removing "module" per the recommendation [here](https://github.com/AnomalyInnovations/serverless-bundle/issues/124#issuecomment-716603300)

Serverless bundle doesn't support CommonJs
https://github.com/AnomalyInnovations/serverless-bundle#module-and-targets

auditLogMover/package.json
```diff
@@ -20,7 +20,7 @@
     "eslint-plugin-prettier": "^3.1.2",
     "jest": "^26.6.3",
     "serverless": "^1.73.1",
-    "serverless-plugin-typescript": "^1.1.9",
```
rsmayda, line 23: Serverless with compile our TS to JS now without this plugin?

auditLogMover/tsconfig.json
rsmayda: Does this need to be propagated to our other packages?

carvantes (dismissed):

scripts/win_install.ps1
```diff
@@ -88,7 +87,9 @@ function Install-Dependencies {
     }
     Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $newpath
     Refresh-Environment
-    
+
+    yarn global add serverless@2.29.0
```
carvantes: Installing `serverless` globally may not be necessary. This runs the risk that the serverless installed globally does not match the version installed locally per `package.json`

This can be a separate change though.

rsmayda: Does this line actually need to be here since it is part of the dev dependencies yarn install should install it, right?

serverless.yaml
```diff
@@ -18,6 +18,10 @@ custom:
   region: ${opt:region, self:provider.region}
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
+  bundle:
```
rsmayda, line 21: In the options https://github.com/AnomalyInnovations/serverless-bundle#options it looks like we may want to specify `packager: yarn` thoughts?

nguyen102: yep, good idea. I'll change it to that

auditLogMover/package.json
nguyen102: Yep, `serverless-bundle` now compiles `TS` for us
https://www.npmjs.com/package/serverless-bundle

scripts/win_install.ps1
nguyen102: Updating to use the `serverless` package that is locally installed and removing global `severless` installation.

nguyen102: Updating to use the `serverless` package that is locally installed and removing global `severless` installation.

auditLogMover/tsconfig.json
nguyen102: Our other packages as in `persistence`, `interface`, etc...? 

In my testing no changes needed to be made to the other packages for serverless deploy to work correctly. My IDE also behave as expected. I also tested linking the deployment package to local version of the routing package. I made code changes and the changes was reflected back and ran successfully.

rsmayda: Awesome thanks

rsmayda (approved):

auditLogMover/package.json
```diff
@@ -20,7 +20,7 @@
...
+    "serverless-bundle": "4.0.1",
```
nguyen102, line 23: Using the latest version of `serverless-bundle`, version `4.2.2` result in this error.

Error: Failed to load config "medikoo" to extend from.
More details here: https://github.com/AnomalyInnovations/serverless-bundle/issues/212

Using version 4.0.1 is the latest version in which that bug doesn't appear

PR #243, opened 2021-03-11
fix: http errors from routing dependency

nguyen102:
Issue #, if available:

Description of changes:
When errors are thrown, the routing package is not catching it properly, hence the failing tests here.
https://github.com/awslabs/fhir-works-on-aws-deployment/runs/2083256307

Errors as shown in CWL
```
ERROR	Unhandled Error TypeError: http_errors_1.default.isHttpError is not a function
    at httpErrorHandler (/var/task/src/webpack:/Volumes/unix/workplace/fhir-works-on-aws-deployment/node_modules/fhir-works-on-aws-routing/lib/router/routes/errorHandling.js:51:1)
```

New changes were testes against crucible and custom integration tests. These tests passed in developer deployed environment. 

Checklist:

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -19,6 +19,8 @@ custom:
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
   bundle:
+    externals:
+      - fhir-works-on-aws-routing
     packager: yarn
     copyFiles: # Copy any additional files to the generated package
       - from: 'bulkExport/glueScripts/export-script.py'


serverless.yaml
```diff
@@ -19,6 +19,8 @@ custom:
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
   bundle:
+    externals:
+      - fhir-works-on-aws-routing
```
rsmayda, line 23: Can you explain how this fixes the issue? and if I were to throw an error in the authz package do we need to add it here as well?

nguyen102: The  method `isHttpError` in the `http-errors` package that is being used in the routing package is not compatible with webpack. Therefore this `external` setting does not use webpack to package the routing package. It use the routing package as is. This setting is defined [here](https://www.npmjs.com/package/serverless-bundle#externals).

RBAC tests was run, and they ran successfully.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/78db6f67101ee6f87365631c33ad88c3da0ef934/integration-tests/rbac-permission.test.ts

These tests involved the authZ package throwing unauthorized errors.

Manually testing creating a Patient as an auditor gives this 403 error, which is expected.
```
{
    "resourceType": "OperationOutcome",
    "text": {
        "status": "generated",
        "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><h1>Operation Outcome</h1><table border=\"0\"><tr><td style=\"font-weight: bold;\">error</td><td>[]</td><td><pre>Unauthorized</pre></td></tr></table></div>"
    },
    "issue": [
        {
            "severity": "error",
            "code": "security",
            "diagnostics": "Unauthorized"
        }
    ]
}
```

rsmayda: What makes a package "not compatible" with webpack? Just trying to understand when we need to add packages to this section

rsmayda: Talked offline -- it is hard to determine if a package is compatible with webpack or no. Having integration tests is our mechanism to prevent these failures

rsmayda (approved):

PR #250, opened 2021-03-15
docs: add Implementation Guides documentation

carvantes:
Description of changes:
Adding Docs for Implementation Guides

### DIFF
--- a/USING_IMPLEMENTATION_GUIDES.md
+++ b/USING_IMPLEMENTATION_GUIDES.md
@@ -1,0 +1,66 @@
+# Using FHIR Implementation Guides
+
+FHIR is a platform specification. It describes a general set of capabilities that can be used to solve many healthcare data exchange problems.
+
+FHIR Implementation guides (IG) are used to describe how FHIR is used in a particular context. IGs are distributed as [packages similar to NPM packages](https://confluence.hl7.org/display/FHIR/NPM+Package+Specification)
+
+## Prerequisites
+
+In addition to the prerequisites described on the [INSTALL docs](INSTALL.md), you need the following: 
+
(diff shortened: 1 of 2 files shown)


USING_IMPLEMENTATION_GUIDES.md
```diff
@@ -0,0 +1,66 @@
...
+1. Java 8 or higher. We recommend using [Amazon Corretto](https://aws.amazon.com/corretto/)
+2. Maven. https://maven.apache.org/install.html
+
+## Installation Steps
+
+1. Obtain the IG packages that you wish to use.
+    
+    IG packages can be obtained from different sources. The most common sources are the corresponding official IG website (i.e. the [US Core website](https://www.hl7.org/fhir/us/core/downloads.html))
+    or from the [FHIR Package Registry](https://registry.fhir.org/)
+
```
nguyen102, line 20: I found it useful to have this command be an example of how to download and unzip a package. 

```
mkdir -p implementationGuides
curl https://www.hl7.org/fhir/us/core/package.tgz | tar xz -C implementationGuides
```

If having the command above is too detailed, then maybe we should just add the direct URL that they can use to get the files that are needed.
https://www.hl7.org/fhir/us/core/package.tgz 

When I just wanted to test IGs, I find this made it a lot easier to get started.

carvantes: Makes sense. 
I agree that the curl command is too specific and I think most customers will download it using  a browser. I'll add the direct link to the download instead.

fyi, this PR will be discarded in favor of the one Yogesh put together, but I'll make sure that this change is reflected there.

carvantes:
Closing in favor of https://github.com/awslabs/fhir-works-on-aws-deployment/pull/256

PR #251, opened 2021-03-15
chore: add third party attributions file

carvantes:
Description of changes:
I crafted this one by hand since I couldn't find a way to automatically generate it.

There exists a popular [maven license plugin](https://www.mojohaus.org/license-maven-plugin/), but it only generates one line per dependency like:
```
(Apache Software License 2.0) HAPI FHIR - Core Library (ca.uhn.hapi.fhir:hapi-fhir-base:5.3.0 - http://jamesagnew.github.io/hapi-fhir/)
(Apache License, Version 2.0) AWS SDK for Java - Core (com.amazonaws:aws-java-sdk-core:1.11.398 - https://aws.amazon.com/sdkforjava)
...
```

### DIFF
--- a/javaHapiValidatorLambda/THIRD-PARTY
+++ b/javaHapiValidatorLambda/THIRD-PARTY
@@ -1,0 +1,1501 @@
+aws-lambda-java-core, aws-lambda-java-events, aws-lambda-java-log4j2
+<https://github.com/aws/aws-lambda-java-libs>
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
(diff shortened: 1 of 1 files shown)


javaHapiValidatorLambda/THIRD-PARTY
```diff
@@ -0,0 +1,1501 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
nguyen102, line 1489: Does this part need to be filled in?
Similar to how this was filled out here
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/feat-search-igs/THIRD-PARTY#L192

carvantes: Not by us. The author of the library should probably do it, but we are just copying their license files as is. I think it is important to not modify in any way their license files.

https://github.com/apache/commons-codec/blob/master/LICENSE.txt#L190

nguyen102 (approved):

PR #253, opened 2021-03-17
chore: add encryption, https-only, access logging to all buckets

carvantes:
Issue #, if available:

Description of changes:
Add encryption, https-only, access logging to all buckets

#### Testing
- Ran integ tests and to ensure that there are no issues with bulk export
- Manually verified that s3 access logs are being written

Checklist:

### DIFF
--- a/cloudformation/bulkExport.yaml
+++ b/cloudformation/bulkExport.yaml
@@ -42,6 +42,33 @@ Resources:
 
   GlueScriptsBucket:
     Type: AWS::S3::Bucket
+    Properties:
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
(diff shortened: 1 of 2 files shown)


cloudformation/bulkExport.yaml
```diff
@@ -42,6 +42,33 @@ Resources:
 
   GlueScriptsBucket:
     Type: AWS::S3::Bucket
+    Properties:
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
```
rsmayda, line 47: Do we want to use the s3 KMS key?

carvantes: default encryption is sufficient for a bucket that only has a copy of the python scripts from github. Also, using KMS with Glue has some added complexity in how the Glue Job permissions are managed.

rsmayda (approved):

PR #254, opened 2021-03-18
feat: add support for DSTU3 on Java Hapi validator

carvantes:
Description of changes:

To use STU3 customers need to:
```
# manually un-comment the Dependencies for FHIR DSTU3 in pom.xml
mvn clean install
serverless deploy --fhirVersion '3.0.1'
```

R4 is the default setting and requires no manual changes:
```
mvn clean install
serverless deploy
```

Also added us-core for r4 and stu3 as resources for unit tests only. The git diff is big but there's no need to review those files.

Checklist:


javaHapiValidatorLambda/pom.xml
```diff
@@ -25,6 +25,63 @@
...
+            <artifactId>hapi-fhir-structures-dstu3</artifactId>
+            <version>5.3.0</version>
+            <!--Note that the scope is test-->
+            <scope>test</scope>
+        </dependency>
+        <!-- ================================================== -->
+
+        <!-- ============ Dependencies for FHIR DSTU3 ============ -->
+<!--
+        <dependency>
```
rsmayda, line 60: So to switch between STU3 and R4 it will be a manual pom.xml change?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -156,6 +199,55 @@ private PrePopulatedValidationSupport loadIgs(final FhirContext ctx) {
             throw new RuntimeException(e);
         }
 
-        return prePopulatedValidationSupport;
+        return new PrePopulatedValidationSupport(ctx, myStructureDefinitions, myValueSets, myCodeSystems);
+    }
+
+    // The following methods are mostly identical as the ones within PrePopulatedValidationSupport.
+    // There is an issue that prevents using PrePopulatedValidationSupport.addValueSet for dstu3 https://github.com/hapifhir/hapi-fhir/issues/2432
```
rsmayda, line 206: +1 thx for opening issue

rsmayda (approved):
LGTM!

PR #274, opened 2021-03-30
chore: Update routing package dependency. Update custom agent env varriable with correct package version

nguyen102:
Issue #, if available:

Description of changes:
chore: Update routing package dependency. Update custom agent env varriable with correct package version

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -49,6 +49,7 @@ provider:
     PATIENT_PICKER_ENDPOINT: ${self:custom.patientPickerEndpoint}
     EXPORT_RESULTS_BUCKET: !Ref BulkExportResultsBucket
     EXPORT_RESULTS_SIGNER_ROLE_ARN: !GetAtt ExportResultsSignerRole.Arn
+    CUSTOM_USER_AGENT: 'AwsLabs/SO0128/1.1.0-smart'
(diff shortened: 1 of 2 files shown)


serverless.yaml
```diff
@@ -49,6 +49,7 @@ provider:
     PATIENT_PICKER_ENDPOINT: ${self:custom.patientPickerEndpoint}
     EXPORT_RESULTS_BUCKET: !Ref BulkExportResultsBucket
     EXPORT_RESULTS_SIGNER_ROLE_ARN: !GetAtt ExportResultsSignerRole.Arn
+    CUSTOM_USER_AGENT: 'AwsLabs/SO0128/1.1.0-smart'
```
rsmayda, line 52: For these user agents is it okay if we use awslabs and not aws solutions?

If not we can add labs to the versionid portion

nguyen102: Yep, it's ok to use `AwsLabs`. I was able to confirm that the metrics appear in the Redshift cluster with `AwsLabs`. In this case there are already metrics in Redshift for when we made API calls using `AWSLABS/SO0128/2.4.0`

The user agent gets capitalized during ETL processing.

rsmayda (approved):

carvantes (approved):

PR #277, opened 2021-04-01
fix: fix unable to locate compiledImplementationGuides/* glob

carvantes:
Add an empty file to ensure that the `compiledImplementationGuides` folder is commited to git

Note: the updates to `yarn.lock` are changes that were missing from an earlier update to `package.json` 

Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,5 +28,6 @@ serverless_config.json
 yalc.lock
 
 /implementationGuides
-/compiledImplementationGuides
+/compiledImplementationGuides/*
+!/compiledImplementationGuides/gitkeep
 /.webpack/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -28,5 +28,6 @@ serverless_config.json
 yalc.lock
 
 /implementationGuides
-/compiledImplementationGuides
+/compiledImplementationGuides/*
+!/compiledImplementationGuides/gitkeep
```
nguyen102, line 32: Nitpick: Should this be `.gitkeep`. That's how I normally see the file added. Technically, I think `gitkeep` would work too, but I would prefer `.gitkeep`

Benefit of `.gitkeep` is it wouldn't normally show up when opening that directory unless someone decides to show hidden files. That reduces the chance of the someone deleting the file.

carvantes: .gitkeep won’t work in this particular case. It would keep the folder in  git but the copy files glob would still fail since glob does not match hidden files

nguyen102 (approved):

PR #285, opened 2021-04-09
test: add token search integ tests

carvantes:
token search integ tests

Checklist:

### DIFF
--- a/integration-tests/search.test.ts
+++ b/integration-tests/search.test.ts
@@ -117,6 +117,46 @@ describe('search', () => {
         }
     });
 
+    test('tokens', async () => {
+        const randomPatientData = randomPatient();
+        randomPatientData.identifier = [
+            {
+                system: 'http://fwoa-integ-tests.com',
+                value: 'someCode',
+            },
+        ];
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', randomPatientData)).data;
+
+        const randomPatientDataNoSystem = randomPatient();
+        randomPatientData.identifier = [
+            {
+                value: 'someCodeWithoutSystem',
+            },
+        ];
+        const testPatientNoSystem: ReturnType<typeof randomPatient> = (
+            await client.post('Patient', randomPatientDataNoSystem)
+        ).data;
+
+        // wait for the patient to be asynchronously written to ES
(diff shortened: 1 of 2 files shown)


integration-tests/search.test.ts
```diff
@@ -117,6 +117,46 @@ describe('search', () => {
...
+        await waitForPatientToBeSearchable(client, testPatient);
+        await waitForPatientToBeSearchable(client, testPatientNoSystem);
+
+        const aFewMinutesAgo = new Date(Date.now() - 1000 * 60 * 10).toJSON();
+        const p = (params: any) => ({ url: 'Patient', params: { _lastUpdated: `ge${aFewMinutesAgo}`, ...params } });
+
+        const testsParamsThatMatch = [
+            p({ identifier: 'http://fwoa-integ-tests.com|someCode' }),
+            p({ identifier: 'someCode' }),
+            p({ identifier: 'http://fwoa-integ-tests.com|' }),
```
rsmayda, line 150: Would the empty ('') search work here too?

integration-tests/search.test.ts
```diff
@@ -117,6 +117,46 @@ describe('search', () => {
...
+                system: 'http://fwoa-integ-tests.com',
+                value: 'someCode',
+            },
+        ];
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', randomPatientData)).data;
+
+        const randomPatientDataNoSystem = randomPatient();
+        randomPatientData.identifier = [
+            {
+                value: 'someCodeWithoutSystem',
```
rsmayda, line 133: Which test cases below test this insert?

integration-tests/search.test.ts
carvantes: It is a valid query, so you get a 200 response with an empty bundle.  But it would fail the test since it doesn't match any patients

integration-tests/search.test.ts
carvantes: it is used on the test on L157. but there was an error on the var name, it should be `randomPatientDataNoSystem.identifier`. It's fixed now

integration-tests/search.test.ts
rsmayda: got it thanks

rsmayda (approved):

PR #291, opened 2021-04-20
feat: enhance numeric and quantity search

carvantes:
Description of changes:
Implementation was done on https://github.com/awslabs/fhir-works-on-aws-search-es/pull/58

Here we are only bumping the `fhir-works-on-aws-search-es` version and adding integ tests

Checklist:

### DIFF
--- a/integration-tests/search.test.ts
+++ b/integration-tests/search.test.ts
@@ -5,6 +5,7 @@
 import { AxiosInstance } from 'axios';
 import waitForExpect from 'wait-for-expect';
 import {
+    aFewMinutesAgoAsDate,
     expectResourceToBePartOfSearchResults,
     expectResourceToNotBePartOfSearchResults,
     getFhirClient,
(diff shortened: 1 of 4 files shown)


integration-tests/search.test.ts
```diff
@@ -161,6 +162,99 @@ describe('search', () => {
...
+            factorOverride: 0.8,
+        };
+
+        const testChargeItem = (await client.post('ChargeItem', chargeItem)).data;
+        await waitForResourceToBeSearchable(client, testChargeItem);
+
+        const aFewMinutesAgo = aFewMinutesAgoAsDate();
+
+        const testsParams = [
+            { 'factor-override': '0.8' },
```
rsmayda, line 238: Is there a use case for quantity being off by a super small amount? 0.0000000005 or something

I thought I remember seeing it in the search code

carvantes: I think you are referring to the comment: https://github.com/awslabs/fhir-works-on-aws-search-es/pull/58#discussion_r615175029. The trailing decimals are rounding errors.

FHIR does equality search by searching a range around the number to work around the possible rounding errors in float numbers. i.e. `eq0.8` or simply `0.8` is effectively searching the range `[0.75, 0.85]` this allows it to match things like `0.8000005`.

There is one test for the range behavior: `eq1.8e2` which searches for the range `[175-180]`, thus matching `185`

rsmayda: Ahh okay thank you

rsmayda (approved):

PR #295, opened 2021-04-22
feat: add DLQ for ddbToEs sync failures

rsmayda:
Issue #, if available: https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/18 & https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/issues/18

Description of changes:
- Add DLQ
- Add Alarms for DLQ
- Add Alarms Lambda IteratorAge

Checklist:

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -115,7 +115,7 @@ functions:
       - stream:
           type: dynamodb
           arn: !GetAtt ResourceDynamoDBTableV2.StreamArn
-          batchSize: 100
+          batchSize: 15 # Lambda payload size limit ~6MB; DDB row limit ~400KB = 15 items
           maximumRetryAttempts: 3
           startingPosition: LATEST
 
@@ -502,6 +502,11 @@ resources:
                       - 'es:ESHttp*'
                     Resource:
                       - !Join ['', [!GetAtt ElasticSearchDomain.Arn, '/*']]
+                  - Effect: Allow
+                    Action:
+                      - 'sqs:SendMessage'
+                    Resource:
+                      - !GetAtt DdbToEsDLQ.Arn
             - PolicyName: 'KMSPolicy'
               PolicyDocument:
                 Version: '2012-10-17'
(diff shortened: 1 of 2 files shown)


serverless.yaml
```diff
@@ -519,7 +524,16 @@ resources:
...
+      DdbToEsDLQ:
+        Type: AWS::SQS::Queue
+        Properties:
+          MessageRetentionPeriod: 1209600 # 14 days in seconds 
   - extensions:
+      DdbToEsEventSourceMappingDynamodbResourceDynamoDBTableV2:
+        Properties:
+          DestinationConfig:
+            OnFailure:
+              Destination: !GetAtt DdbToEsDLQ.Arn
```
carvantes, line 536: Not strictly needed but consider explicitly adding `"MaximumRetryAttempts": 3` here.

Looks like serverless adds that by default, which is a bit weird since the AWS default for streams is to retry an unlimited number of times(until the message expires). 

At first glance this looks like a config error since it's hard to tell that serverless adds `MaximumRetryAttempts`. I think that having `OnFailure` without setting `MaximumRetryAttempts` is effectively useless since Lambda keeps retrying and never "fails", I'm not sure if that config combination is even allowed.

carvantes (approved):
Looks good. Just a minor comment about a change that'd make the config clearer

serverless.yaml
rsmayda: So it is actually being set in the event block in the start: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/06d8db534d0b993d7958f32fcda9a52ff5a4dc60/serverless.yaml#L114-L120

would you suggest adding a comment here instead?

carvantes: nvm, we can leave it as is. 
I didn't realize the stream config was above. A comment may help but I think the Serverless framework `extensions` will aways have the issue that config ends up split into 2 places.

PR #296, opened 2021-04-22
feat: Add post search and integ tests

Bingjiling:
Issue #, if available:

Description of changes:

* Bump up routing package version number to include post search change 
* Add integ tests

Checklist:

### DIFF
--- a/integration-tests/search.test.ts
+++ b/integration-tests/search.test.ts
@@ -68,6 +68,43 @@ describe('search', () => {
         }
     });
 
+    test('search for various valid parameters in query and request body', async () => {
+        const testPatient: ReturnType<typeof randomPatient> = (await client.post('Patient', randomPatient())).data;
+
+        // wait for the patient to be asynchronously written to ES
+        await waitForResourceToBeSearchable(client, testPatient);
+
+        const aFewMinutesAgo = aFewMinutesAgoAsDate();
+
+        const p = (params: any, postQueryParams: any) => ({
+            url: 'Patient',
+            params: { _lastUpdated: `ge${aFewMinutesAgo}`, ...params },
+            postQueryParams,
+        });
+        const testsParams = [
+            p({ 'address-city': testPatient.address[0].city }, { 'address-country': testPatient.address[0].country }),
+            p(
+                { 'address-postalcode': testPatient.address[0].postalCode },
+                { family: testPatient.name[0].family, 'address-postalcode': testPatient.address[0].postalCode },
+            ),
+            p(
+                { name: testPatient.name[0].given[0] },
+                { name: testPatient.name[0].given[0], gender: testPatient.gender },
+            ),
+            p(
+                { phone: testPatient.telecom.find(x => x.system === 'phone')!.value },
+                { email: testPatient.telecom.find(x => x.system === 'email')!.value },
+            ),
+        ];
+
+        // run tests serially for easier debugging and to avoid throttling
+        // eslint-disable-next-line no-restricted-syntax
+        for (const testParams of testsParams) {
+            // eslint-disable-next-line no-await-in-loop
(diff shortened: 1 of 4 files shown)


integration-tests/utils.ts
```diff
@@ -145,16 +146,16 @@ const expectSearchResultsToFulfillExpectation = async (
...
-    ).resolves.toMatchObject({
+    console.log('GET Searching with params:', search);
+    const searchResult = (await client.get(search.url, { params: search.params })).data;
+    expect(searchResult).toMatchObject({
+        resourceType: 'Bundle',
+        entry: bundleEntryExpectation,
+    });
+
+    console.log('POST Searching with params:', search);
+    const postSearchResult = (await client.post(`${search.url}/_search`, qs.stringify(search.params))).data;
```
rsmayda: Could we add a test case of both in query param and in the post body?

rsmayda: Also do we need qs here to stringify? the params would JSON.stringify work?

Bingjiling: I tried JSON.stringify first, it didn't work unfortunately.

carvantes: Aren't you using `qs` to have `Content-Type=application/x-www-form-urlencoded`?

integration-tests/utils.ts
```diff
@@ -145,16 +146,25 @@ const expectSearchResultsToFulfillExpectation = async (
...
+    expect(postSearchResult).toMatchObject({
+        resourceType: 'Bundle',
+        entry: bundleEntryExpectation,
+    });
+
+    console.log('POST Searching with same params in body and in query:', search);
+    const postSearchRepeatingParamsResult = (
+        await client.post(`${search.url}/_search`, qs.stringify(search.params), { params: search.params })
+    ).data;
+    expect(postSearchRepeatingParamsResult).toMatchObject({
```
carvantes: I think that we should write this as a separate test case.

Running this for ALL test cases will be slow. Plus, this is not an equivalent search request to the ones above. Duplicating params MAY have a different meaning. Doing POST for ALL test cases is already doubling the execution time for this tests, but that's probably ok for now.

I think it's better to just add a couple test cases that send some query params + some params in the body.

Bingjiling: Emm, yea that makes sense. I'll update it. 

(Out of curiosity, why were you up and giving comments 4hrs ago? That's awfully early for you. )

integration-tests/utils.ts
Bingjiling: JSON.stringify and qs.stringify do different things actually. If we take `{ x: 5, y: 6 }` as example, JSON.stringify would give `"{"x":5,"y":6}"`, and qs.stringify would give `"x=5&y=6"`.

integration-tests/search.test.ts
```diff
@@ -68,6 +68,43 @@ describe('search', () => {
...
+        });
+        const testsParams = [
+            p({ 'address-city': testPatient.address[0].city }, { 'address-country': testPatient.address[0].country }),
+            p(
+                { 'address-postalcode': testPatient.address[0].postalCode },
+                { family: testPatient.name[0].family, 'address-postalcode': testPatient.address[0].postalCode },
+            ),
+            p(
+                { name: testPatient.name[0].given[0] },
+                { name: testPatient.name[0].given[0], gender: testPatient.gender },
```
rsmayda, line 92: How does the combination work is it an AND or an OR? i.e. if we have name = rob then name = bob what would the result be an OR or an AND?

Bingjiling: It means AND. If we have name = rob and name = bob it would result in name=[rob, bob] and pass to search, it'll be up for search component to decide what it means.

rsmayda (approved):

PR #298, opened 2021-04-23
chore: lock down versions of external deployment dependencies

nguyen102:
Description of changes:
Lock down `serverless` and `yarn` version. Serverless is locked down because we're now using the locally installed version of serverless. That version is defined in our `yarn.lock` file.

For `yarn` we are explicitly downloading version `1.22.5`

These script was tested and passed on the following environments
* Ubuntu
* Amazon Linux
* Mac 
* Windows

Checklist:

### DIFF
--- a/scripts/win_install.ps1
+++ b/scripts/win_install.ps1
@@ -69,7 +69,6 @@ function Install-Dependencies {
     }
     choco install -y nodejs.install --version=12.18.3 #also installs npm by default
     choco install -y python3
-#     choco install -y yarn
     npm install --global yarn@1.22.5
 
     #fix path issues
--- a/scripts/win_install.ps1
+++ b/scripts/win_install.ps1
@@ -69,7 +69,8 @@ function Install-Dependencies {
     }
     choco install -y nodejs.install --version=12.18.3 #also installs npm by default
     choco install -y python3
-    choco install -y yarn
+#     choco install -y yarn
+    npm install --global yarn@1.22.5
 
     #fix path issues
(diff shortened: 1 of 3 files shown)


serverless.yaml
```diff
@@ -22,8 +22,8 @@ custom:
     copyFiles: # Copy any additional files to the generated package
       - from: 'bulkExport/glueScripts/export-script.py'
         to: './bulkExport/glueScripts/export-script.py'
-      - from: 'compiledImplementationGuides/*'
-        to: './'
+      - from: 'compiledImplementationGuides'
+        to: './compiledImplementationGuides'
```
nguyen102, line 26: This change was made because Windows does not like the `*` wildcard. Error shown below
```
ERROR in unable to locate 'C:\Users\Administrator\Documents\fhir-works-on-aws-deployment\compiledImplementationGuides\*'
 glob
```

scripts/win_install.ps1
```diff
@@ -69,7 +69,8 @@ function Install-Dependencies {
     }
     choco install -y nodejs.install --version=12.18.3 #also installs npm by default
     choco install -y python3
-    choco install -y yarn
+#     choco install -y yarn
```
carvantes: Nit: delete instead of commenting out

carvantes (approved):

scripts/win_install.ps1
nguyen102: ooops, deleting now.

PR #310, opened 2021-05-03
feat: deployment change for adding logging framework 

Bingjiling:
Issue #, if available:

Description of changes:

* Deployed and tested in AWS account, all integ tests succeed 

Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,9 @@
     "aws-sdk": "^2.785.0",
     "axios": "^0.21.1",
     "fhir-works-on-aws-authz-rbac": "4.1.1",
-    "fhir-works-on-aws-interface": "8.1.1",
+    "fhir-works-on-aws-interface": "./.yalc/fhir-works-on-aws-interface",
     "fhir-works-on-aws-persistence-ddb": "3.3.3",
-    "fhir-works-on-aws-routing": "5.2.0",
+    "fhir-works-on-aws-routing": "./.yalc/fhir-works-on-aws-routing",
     "fhir-works-on-aws-search-es": "2.5.1",
     "serverless-http": "^2.3.1",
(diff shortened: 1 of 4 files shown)


serverless.yaml
```diff
@@ -15,6 +15,11 @@ custom:
   oauthRedirect: ${opt:oauthRedirect, 'http://localhost'}
   config: ${file(serverless_config.json)}
   useHapiValidator: ${opt:useHapiValidator, 'false'}
+  logLevel:
+    !If
+    - isDev
+    - 'debug'
+    - 'info'
```
rsmayda:
```suggestion
  useHapiValidator: ${opt:useHapiValidator, 'false'}
  logLevel: ${opt:logLevel, 'error'}
```

I would suggest making it deployer's choice but default to the cheapest option ('error'). We will also need to handle the case when they input an invalid level

package.json
```diff
@@ -39,9 +39,9 @@
     "aws-sdk": "^2.785.0",
     "axios": "^0.21.1",
     "fhir-works-on-aws-authz-rbac": "4.1.1",
-    "fhir-works-on-aws-interface": "8.1.1",
-    "fhir-works-on-aws-persistence-ddb": "3.3.3",
-    "fhir-works-on-aws-routing": "5.2.0",
+    "fhir-works-on-aws-interface": "9.0.0",
+    "fhir-works-on-aws-persistence-ddb": "3.3.4",
+    "fhir-works-on-aws-routing": "5.3.0",
     "fhir-works-on-aws-search-es": "2.5.1",
```
carvantes: looks like we are we missing the search update

Bingjiling: Good catch!

carvantes (approved):

PR #312, opened 2021-05-04
!fix: make glue IAM role more restrictive

rsmayda:
### Description of changes:

Due to Glue IAM `Resource` definition for the required job actions the `Resource` needs to be `*`. To better lock down the access of this IAM role, it is advised to add a tag to the Glue job then add a condition to the IAM Role.

Good example of this provided by AWS: https://docs.aws.amazon.com/glue/latest/dg/glue-policy-examples-iam-tags.html

NOTE: this is a breaking change due to: https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/306
- Existing FWoA deployments will need to manually add the tag to the existing Glue job: `Name: !Sub '${AWS::StackName}-ExportGlueJob'`

### Checklist:

* [x ] Have you successfully deployed to an AWS account with your changes?

### DIFF
--- a/cloudformation/bulkExport.yaml
+++ b/cloudformation/bulkExport.yaml
@@ -33,6 +33,9 @@ Resources:
                   - glue:GetJobRun
                   - glue:BatchStopJobRun
                 Resource: '*'
+                Condition:
+                  StringEquals:
+                    'glue:resourceTag/Name': !Sub '${AWS::StackName}-ExportGlueJob'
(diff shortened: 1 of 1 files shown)


nguyen102 (approved):

rsmayda:
Strictly speaking, this is undocumented policy behavior. We do not want to take on the additional risk of this change.

PR #326, opened 2021-05-13
feat: update IGCompiler to accept OperationDefinition

carvantes:
just reading `OperationDefinition` files and passing them to the routing compiler

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-routing/pull/83

Checklist:

### DIFF
--- a/src/implementationGuides/IGCompiler.ts
+++ b/src/implementationGuides/IGCompiler.ts
@@ -56,21 +56,21 @@ export class IGCompiler {
 
     private readonly searchImplementationGuides: ImplementationGuides;
 
-    private readonly structureDefinitionImplementationGuides: ImplementationGuides;
+    private readonly routingImplementationGuides: ImplementationGuides;
 
     constructor(
         searchImplementationGuides: ImplementationGuides,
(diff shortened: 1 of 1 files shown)


src/implementationGuides/IGCompiler.ts
```diff
@@ -79,7 +79,7 @@ export class IGCompiler {
         const index: any = await loadJson(indexJson);
         const resources = [];
         for (const file of index.files) {
-            if (file.resourceType === resourceType) {
+            if (resourceTypes.includes(file.resourceType)) {
```
Bingjiling, line 82: Would this evaluate to true if `resourceTypes = ['SearchParameter', 'StructureDefinition']` and `resourceType = ['SearchParameter', 'StructureDefinition']` ?

Bingjiling (approved):
LGTM, one small question.

src/implementationGuides/IGCompiler.ts
carvantes: It wouldn't, but that scenario cannot happen.
`file.resourceType` is always a string, not an array of strings.

PR #346, opened 2021-06-10
chore: script for add alias to existing index

Bingjiling:
Issue #, if available:

Description of changes:

* Add script for adding aliases to existing indices

Checklist:

### DIFF
--- a/scripts/elasticsearch-operations.js
+++ b/scripts/elasticsearch-operations.js
@@ -1,4 +1,4 @@
-/* 
+/*
  *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *  SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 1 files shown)


scripts/elasticsearch-operations.js
```diff
@@ -110,6 +110,43 @@ const deleteIndex = async () => {
     }
 };
 
+const addAlias = async () => {
```
carvantes, line 113: I'd suggest adding a comment/docs describing what this function does.

I know that the rest of this file has almost no documentation, but I feel like this one deserves it since It will be mentioned as the path to upgrade for breaking change on the release notes.

All the other functions are extras that I'm not sure if they are ever used (I didn't even know this file existed).

carvantes (approved):

scripts/elasticsearch-operations.js
Bingjiling: Good point! I'll add documentation for it in subsequent PR. I was not aware of this file exited before either, I'll bring it up in tmr's standup to see if we still want to keep them.

PR #347, opened 2021-06-11
feat: remove unneeded scope checks in authorizer

carvantes:
Description of changes:

Checking for the `openid`, `profile`, `aws.cognito.signin.user.admin` scopes on the APIGW authorizer is not really doing anything for us. Our AuthZ is entirely based on the `cognito:groups`

This feature is mostly used with custom scopes, and when the API methods are all defined in APIGW(unlike our big `{proxy+}` method) so that different methods are authorized by different scopes. [e.g.](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-define-resource-servers.html#cognito-user-pools-define-resource-servers-console) creating a custom scope `weather.read` to be used on a weather API

btw, this change simplifies the multi-tenancy setup.

Checklist:

* [n/a] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,9 +117,15 @@ To find what FHIR Server supports, use the `GET Metadata` Postman request to ret
 
 **Authorizing a user**
 
-FHIR Works on AWS uses Role-Based Access Control (RBAC) to determine what operations and what resource types a user can access. The default rule set can be found in [RBACRules.ts](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/src/RBACRules.ts). The role of an user is determined from the Cognito group that they belong to. To access the API, you must use the OAuth access token. 
+FHIR Works on AWS uses Role-Based Access Control (RBAC) to determine what operations and what resource types a user can access. The default rule set can be found in [RBACRules.ts](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/src/RBACRules.ts). To access the API, you must use the OAuth access token. This access token must include scopes of either `openid`, `profile` or `aws.cognito.signin.user.admin`. 
 
-For more information, click [here](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-user-groups.html).
+Using either of these scopes provide information about users and their group. It helps determine what resources/records they can access.
+
++ The `openid` scope returns all user attributes in the ID token that are readable by the client. The ID token is not returned if the openid scope is not requested by the client. 
++ The `profile` scope grants access to all user attributes that are readable by the client. This scope can only be requested with the openid scope. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,15 +117,9 @@ To find what FHIR Server supports, use the `GET Metadata` Postman request to ret
 
 **Authorizing a user**
 
-FHIR Works on AWS uses Role-Based Access Control (RBAC) to determine what operations and what resource types a user can access. The default rule set can be found in [RBACRules.ts](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/src/RBACRules.ts). To access the API, you must use the OAuth access token. This access token must include scopes of either `openid`, `profile` or `aws.cognito.signin.user.admin`.
```
rsmayda: To get groups to be put into the JWT the requester must use these scopes during the request (this was required about a year ago perhaps it changed).

So the idea behind these is the fail early principal (fail at authorized level and not even invoke proxy lambda). That said if removing simplifies the flow I am okay removing it.

rsmayda: Looking at the RBAC package if groups is null in JWT token it will be an empty array and will result in "Unauthorized" exception

rsmayda (approved):
Nothing blocking; leaving the explicit scopes needed in the README maybe helpful, but you are linking to the cognito docs

rsmayda (changes requested):
I double checked the docs AND I couldn't find where it explicitly states how to get group claim in the access token. Please find a better link or leave in the required scopes in README

carvantes:
> I double checked the docs AND I couldn't find where it explicitly states how to get group claim in the access token. Please find a better link or leave in the required scopes in README

ok, I reverted the docs changes. I didn't realize that `cognito:groups` may not be present depending of scopes.

I couldn't find concise docs either, but from my tests I found that `openId` scope is sufficient for `cognito:groups` to show up, and that  doing `initiate_auth` with the SDK always returns `cognito:groups` regardless of scopes. Still, I think it's ok to leave all the scopes in the docs since we are not changing the  [`client AllowedOAuthScopes`](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/94f0598b5125cd3d76f00ac89a452208b0dc804b/cloudformation/cognito.yaml#L32)  

```json
{
  "sub": "xxxxxxxxxx",
  "cognito:groups": [
    "practitioner"
  ],
  "token_use": "access",
  "scope": "openid",
  "auth_time": 1623428796,
  "iss": "https://cognito-idp.us-east-2.amazonaws.com/us-east-2_xxxxxxxx",
  "exp": 1623432396,
  "iat": 1623428796,
  "version": 2,
  "jti": "8cbfa15e-a64b-44c1-bac0-a03214e56e0f",
  "client_id": "xxxxxxxxxx",
  "username": "tenant33345643"
}
```

rsmayda (approved):

PR #348, opened 2021-06-11
feat: add tenantId attribute to Cognito user pool

carvantes:
Description of changes:
Adding tenantId as an attribute to user pool.

Also, minor updates to py scripts so that they print both the `AccessToken` and the `IdToken`. Multitenancy requires the use of the `IdToken` since it is impossible to add custom claims on the `AccessToken`. Other than custom claims, both tokens are interchangeable in our case, since `cognito:groups` is present on both tokens and the Cognito user pools APIGW Authorizer accepts Id and access tokens indistinctively.

Checklist:

* [n/a ] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/cloudformation/cognito.yaml
+++ b/cloudformation/cognito.yaml
@@ -22,6 +22,8 @@ Resources:
           Required: true
         - AttributeDataType: String
           Name: cc_confirmed
+        - AttributeDataType: String
+          Name: tenantId
   UserPoolClient:
     Type: AWS::Cognito::UserPoolClient
     Properties:
--- a/scripts/init-auth.py
+++ b/scripts/init-auth.py
@@ -4,6 +4,7 @@
 """
 import boto3
 import sys
+import json
 
 client = boto3.client('cognito-idp', region_name=sys.argv[2])
 '''
@@ -21,5 +22,4 @@ response = client.initiate_auth(
     ClientId=sys.argv[1]
 )
 
-sessionid = response['AuthenticationResult']['AccessToken']
-print(sessionid)
+print(json.dumps(response['AuthenticationResult'], indent=2))
--- a/scripts/provision-user.py
+++ b/scripts/provision-user.py
@@ -5,7 +5,7 @@
 
 import boto3
 import sys
-
+import json
 
 '''
 example run:
@@ -15,9 +15,11 @@ python3 provision-user.py us-west-2_yk8jbgpWM 12pgvi3gsl32qp9h8lg130arr0 us-west
 
 client = boto3.client('cognito-idp', region_name=sys.argv[3])
 
+USERNAME = 'workshopuser'
+
 response = client.admin_create_user(
     UserPoolId=sys.argv[1],
-    Username='workshopuser',
+    Username=USERNAME,
     UserAttributes=[
         {
             'Name': 'email',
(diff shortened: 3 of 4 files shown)


rsmayda:
Have we considered getting custom attributes from Cognito via api calls? I am not sure how common it is for providers to pass in Id token instead of access

Auth0 is pretty opinionated about it: "Do not use ID tokens to gain access to an API."

https://auth0.com/docs/tokens

carvantes:
@rsmayda 
> Have we considered getting custom attributes from Cognito via api calls? I am not sure how common it is for providers to pass in Id token instead of access
> 
> Auth0 is pretty opinionated about it: "Do not use ID tokens to gain access to an API."
> 
> https://auth0.com/docs/tokens

For  APIGW + Cognito, using IdToken or AccessToken are both safe and equally valid. The AWS docs never give preference to either one, rather they mention that they work best for different use cases.

[link - Control access to a REST API using Amazon Cognito user pools as authorizer - Amazon API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-integrate-with-cognito.html)
> The identity token is used to authorize API calls based on identity claims of the signed-in user. The access token is used to authorize API calls based on the custom scopes of specified access-protected resources.

It's hard to tell how common it is to use IdTokens. As an example, AWS amplify(a popular tool for building client apps that integrate with AWS services) allows customers to use either one, but actually use idTokens in many of their examples.
https://docs.amplify.aws/lib/restapi/authz/q/platform/js#note-related-to-use-access-token-or-id-token

In general, apps/services that use custom claims favor using IdTokens. e.g. [performance-dashboard-on-aws](https://github.com/awslabs/performance-dashboard-on-aws/blob/d742291d207ce4ba4206a10a4242c0fdfc8d528a/frontend/src/services/BackendService.ts#L23-L27), [AWS blog post about using custom claims for multi-tenancy](https://aws.amazon.com/blogs/apn/managing-saas-identity-through-custom-attributes-and-amazon-cognito/)

> Have we considered getting custom attributes from Cognito via api calls?

We could build such solution. It would be either a custom authorizer or additional code on the RBAC auth package. 

**Use AccessToken + callback to Cognito to fetch custom claims**
Pros:
- It is more OIDC compliant

Cons:
- It is expensive/complex ( +1 Cognito call for ALL requests, need to build caching for performance), 
- It sort of defeats the point of using JWT
- We are not authorizing using scopes anyways (scopes are the OIDC way of doing authZ).  

**Use IdToken** 
Pros: 
- Very simple. The built in Cognito Pools APIGW authorizer does the job.

Cons: 
- Does not strictly follow OIDC
- It is a breaking change since we are asking customers to send IdToken in requests instead of AccessToken. Not a massive change, since any application calling FWoA would have access to both tokens(tokens have same lifecycle, refresh token works for both). They just need to switch to sending the other token as auth headers.
Note: It is not possible to accept both IdTokens and AccessTokens on the same APIGW using the built-in Cognito Pools authorizer. To keep things simple we'd probably switch to IdTokens regardless of whether or not multi-tenancy is enabled.

I’m in favor of using IdTokens here, but we can discuss further.

rsmayda (approved):
Talked on a call approved as is!

PR #367, opened 2021-06-25
feat: update lambda state machine to accommodate tenantId

Bingjiling:
Issue #, if available:

Description of changes:

* State machine and glue script change to accommodate tenantId
* Depends on PR https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/90 

Checklist:

### DIFF
--- a/bulkExport/glueScripts/export-script.py
+++ b/bulkExport/glueScripts/export-script.py
@@ -31,6 +31,9 @@ if ('--{}'.format('type') in sys.argv):
 groupId = None
 if ('--{}'.format('groupId') in sys.argv):
     groupId = getResolvedOptions(sys.argv, ['groupId'])['groupId']
+tenantId = None
+if ('--{}'.format('tenantId') in sys.argv):
+    tenantId = getResolvedOptions(sys.argv, ['tenantId'])['tenantId']
 
 job_id = args['jobId']
 export_type = args['exportType']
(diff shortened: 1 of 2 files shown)


bulkExport/glueScripts/export-script.py
```diff
@@ -60,12 +63,21 @@
...
 )
 
+print('Start filtering by tenantId')
+# Filter by resource listed in Type and with correct STATUS
+if (tenantId is None):
+    filtered_tenant_id_frame = original_data_source_dyn_frame
+else:
+    filtered_tenant_id_frame = Filter.apply(frame = original_data_source_dyn_frame,
+                               f = lambda x:
+                               x['_tenantId'] == tenantId)
```
rsmayda, line 73: instead of doing glue side filtering would it be better to have a secondary index on the tenantId? This will become an expensive operation if we have to scan across all tenants

Bingjiling: Emm that's a great question. In the design doc, it specified the filtering is to be done as part of the Glue job, and secondary index was not introduced for any tables. @carvantes Any thoughts?

carvantes: The glue job always scans the entire DDB table no matter what, there's no way to use a query. This is a limitation on the current AWS Glue + DDB integration.

There are existing scenarios where this is far from ideal. e.g. exporting a single FHIR resource type or exporting the resources modified in the last hour will both scan the entire table.

There is room for improvement on the bulk export solution, but we are not changing the fundamentals here.

bulkExport/glueScripts/export-script.py
```diff
@@ -60,12 +63,21 @@
     }
 )
 
+print('Start filtering by tenantId')
+# Filter by resource listed in Type and with correct STATUS
```
carvantes: update this comment

carvantes (approved):

PR #376, opened 2021-07-01
chore: update postman collection with more resources

Zambonilli:
Issue #, if available:

Description of changes:
Added a script to autogenerate the postman collection from the public FHIR postman collection and HL7 json examples. Updated the postman collection from a run of the script so we have more resources that are able to be ran from postman.

Checklist:

### DIFF
--- a/scripts/create-postman-collection.ts
+++ b/scripts/create-postman-collection.ts
@@ -1,0 +1,173 @@
+'use strict';
+import fs from 'fs';
+import _ from 'lodash';
+import path from 'path';
+
+/*
+    Script to help auto-generate our FWoA postman collection for testing.
+    Postman has a public collection for FHIR with most R4 resources here
+    https://www.postman.com/api-evangelist/workspace/fast-healthcare-interoperability-resources-fhir/overview
+    Examples from HL7 FHIR R4 downloads here
+    https://www.hl7.org/fhir/examples-json.zip
+
+    This script can take an export from the public FHIR collection, and 
+    merge it with our definitions in ../postman
+
+    `npx ts-node create-postman-collection.ts $your-public-collection-file $your-unziped-examples-dir $output-file`
+    EG: npx ts-node ./scripts/create-postman-collection.ts ~/Downloads/FHIR.postman_collection.json ~/Downloads/examples-json /tmp/FHIR.fwoa_postman_collection.json
+*/
+
+const updateRequests = (item:any, auth:any, resourceName:string, examplesDir: string)=> {
+    if (_.has(item, 'request')){
+
+        const hasResponse = _.has(item, 'response') && item.response.length > 0;
+
+        // add the auth
+        item.request.auth = _.cloneDeep(auth);
+
+        // add the x-api-key header
+        if (!_.has(item.request, 'header')){
+            item.request.header = [];
+        }
+        item.request.header.push({
+            key: "x-api-key",
+            value: "{{API_KEY}}",
+            type: "text"
+        });
+        if (hasResponse) {
+            if (_.has(item.response[0], 'header')){
+                item.response[0].header.push({
+                    key: "x-api-key",
+                    value: "{{API_KEY}}",
+                    type: "text"
+                });
+            }
+        }
+
+        // update to {{API_URL}} from {{baseUrl}}
+        item.request.url.raw = item.request.url.raw.replace("{{baseUrl}}", "{{API_URL}}");
+        item.request.url.host = [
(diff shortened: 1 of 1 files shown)


scripts/create-postman-collection.ts
```diff
@@ -0,0 +1,177 @@
...
+
+        // pluck out the auth from an existing route
+        const fwoaAuthItem = _.find(fwoaCollection.item, item => {
+            return _.has(item, 'request') && _.has(item.request, 'auth') && item.request.auth.type !== 'noauth';
+        });
+        console.log('found fwoa auth');
+        if (_.isUndefined(fwoaAuthItem)) {
+            console.log('fwoa postman collection does not have any requests defined');
+            process.exit(1);
+        }
```
rsmayda: I believe we want to log success after this check
```suggestion
        if (_.isUndefined(fwoaAuthItem)) {
            console.log('fwoa postman collection does not have any requests defined');
            process.exit(1);
        }
        console.log('found fwoa auth');
```

scripts/create-postman-collection.ts
```diff
@@ -0,0 +1,177 @@
...
+        console.log('parsing complete');
+
+        if (!_.has(fwoaCollection, 'item') || fwoaCollection.item.length === 0) {
+            console.log('fwoa postman collection does not have any items defined');
+            process.exit(1);
+        }
+
+        // pluck out the auth from an existing route
+        const fwoaAuthItem = _.find(fwoaCollection.item, item => {
+            return _.has(item, 'request') && _.has(item.request, 'auth') && item.request.auth.type !== 'noauth';
```
rsmayda: in TS you can use `?.` - so I think something like this would work: 
```suggestion
            return item?.request?.auth?.type !== 'noauth';
```

Called optional chaining: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

rsmayda: You could do this for any `_.has` calls

scripts/create-postman-collection.ts
```diff
@@ -0,0 +1,177 @@
...
+            });
+            if (_.isUndefined(fwoaItem)) {
+                // new item just add after updates
+                const newItem = _.cloneDeep(item);
+
+                updateRequests(newItem, auth, resourceName, examplesDir);
+
+                fwoaCollection.item.push(newItem);
+            } else {
+                // TODO: go recursive here
```
rsmayda: Just to understand - this would be how do you handle the situations when FWoA and your public source has a resource? perhaps update the comment to be clearer?

rsmayda (changes requested):
Thank you for making this contribution! I was able to upload the configuration and it looks really good.

Some minor changes to make. Could we not include the `examples` (see screenshot), IMO they are not adding any value? Looking at the json I believe that is the `"response"` field in the collection

image

Also I think we should "overwrite" the existing resources so they are all uniform (Patient, ...) and the only custom FWoA ones are Bundles and Metadata

scripts/create-postman-collection.ts
Zambonilli: woah, that's awesome thanks. I really need to break myself of lodash/underscore and start using the native js/ts options.

scripts/create-postman-collection.ts
Zambonilli: Yeah, this would be if both the public collection and fwoa collection have the same top level directory. I didn't want to overwrite the existing requests so there was a thought that we could add any public ones to the same directory. Looks like from upper feedback it's better to just use public + fwoa custom requests so for now they would be one in the same. I can update so if we want to add custom requests to a top level resource we merge public w/fwoa.

scripts/create-postman-collection.ts
Zambonilli: Updated to use public patient and binary request definitions. So, the only remaining definitions from the existing fwoa are bundle, export and metadata.

Updated the success print statement location.

scripts/create-postman-collection.ts
Zambonilli: Switched over to optional chaining, thanks again for the tip!

scripts/create-postman-collection.ts
Zambonilli: better comments left in the TODO here with hints on how to refactor and walk both the fwoa collection and public collection trees to do the merge.

Zambonilli:
Pushed updates for all the feedback provided. Have a great holiday weekend.

rsmayda (approved):

PR #384, opened 2021-07-16
fix: remove _id, _tenantId from bulk export results

carvantes:
remove multi-tenancy related internal fields from bulk export results.

### DIFF
--- a/bulkExport/glueScripts/export-script.py
+++ b/bulkExport/glueScripts/export-script.py
@@ -64,13 +64,21 @@ original_data_source_dyn_frame = glueContext.create_dynamic_frame.from_options(
 )
 
 print('Start filtering by tenantId')
+
+def remove_composite_id(resource):
+  # Replace the multi-tenant composite id with the original resource id found at "_id"
+  resource["id"] = resource["_id"]
+  return resource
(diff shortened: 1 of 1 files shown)


bulkExport/glueScripts/export-script.py
```diff
@@ -96,7 +104,7 @@
 
 # Drop fields that are not needed
 print('Dropping fields that are not needed')
-data_source_cleaned_dyn_frame = DropFields.apply(frame = filtered_dates_resource_dyn_frame, paths = ['documentStatus', 'lockEndTs', 'vid', '_references'])
+data_source_cleaned_dyn_frame = DropFields.apply(frame = filtered_dates_resource_dyn_frame, paths = ['documentStatus', 'lockEndTs', 'vid', '_references', '_tenantId', '_id'])
```
Bingjiling, line 107: Would it throw error or warning if _tenantId and _id does not exist?

carvantes: No, `DropFields` ignores fields that don't exist on the frame (I tested it)

rsmayda (approved):

Issue #378, opened 2021-07-06
[Misc] private and public fhir-service API Gateway


Zambonilli:
**What's on your mind?**
We're looking to have traffic to fwoa routed internally from our ec2 instances. However, when trying to route the traffic internally, we were unable to because the fhir-service is currently public. Setting up a VPC Endpoint for API gateway blocks the public DNS for the fhir-service unless it's a custom DNS. Looking more closely at the documentation for setting up VPC endpoints for API Gateway, the diagrams depict creating both a public and a private API in API Gateway if you want to service both kinds of traffic.

Is this analysis correct or is there a better way to use the existing fhir-service and route internally? If the analysis is correct, is adding a private fhir-service something that we would want as part of fwoa? I can see a lot of implementations wanting both a private and public APIs here.

**Versions (please complete the following information):**
 - v3.0.0

rsmayda:
A private API GW is something of interest for some of our customers. Unfortunately we have not had a chance to dive deep into how or what it would look like. I will add this to our backlog! If you need a quicker answer we can rope in some SA specialists and pick their brains.

Zambonilli:
Thanks, this is blocking us so we'll dive on it and get a PR going.

Zambonilli:
FYI, we have a branch in our fork that is setting up a private API gateway here. https://github.com/Zambonilli/fhir-works-on-aws-deployment/tree/feature/private_api_gateway with most of the change in https://github.com/Zambonilli/fhir-works-on-aws-deployment/blob/feature/private_api_gateway/cloudformation/apiGateway.yaml

We were able to leverage the clone function in API Gateway to clone the public API and change the EndpointType to private. The Deployment and stages are not cloned so we were able to reference the serverless provider settings to match that of the public API. We have to still do a full audit of every field between public and private APIs and regression test our use of FHIR in our application when calling the private API but it looks pretty solid from the early results.

We'll open a PR once we've completed our audit and regression testing.

Zambonilli:
[...]

PR #385, opened 2021-07-19
feat: private api gateway

Zambonilli:
Issue #, if available:
378
https://github.com/awslabs/fhir-works-on-aws-deployment/issues/378

Description of changes:
Added cloudformation file for private API gateway. Most of the infrastructure leverages the API Gateway's clone REST API functionality but stage information and usage plans are created independently.

Checklist:


Zambonilli:
FYI, we hit 1 issue with this implementation in that the API_URL logic being passed throughout the modules expects there to be a single base url in which to build upon. With this PR, there is a public and a private API gateway both with different assigned URLs. In addition, private API gateways have a slew of ways they can be called besides just the private API gateway assigned DNS (https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-api-test-invoke-url.html). We have manually set the API_URL environment variable on the fhir-service lambda to the private DNS to get through. Long term, we'll need to refactor the deployment, routing and persistence modules to handle the potential for different base API urls on each request.

Zambonilli:
pulling this until there's a solution for multiple API_URLs

PR #387, opened 2021-07-21
test: add multi-tenancy integ tests

carvantes:
An env variable must be set to run the integ tests: `MULTI_TENANCY_ENABLED=true yarn int-test`. 
This enables the tests specific to multi-tenancy and also slightly changes how the axios client is built on `getFhirClient` so that all other test suites run successfully.


package.json
```diff
@@ -20,7 +20,7 @@
     "lint-fix": "eslint --fix . --ext .ts,.tsx",
     "build": "tsc",
     "watch": "tsc -w",
-    "test": "jest --silent --passWithNoTests --testPathPattern=src/*",
+    "test": "jest --silent --passWithNoTests --testPathPattern=src/* --runInBand",
```
carvantes: multitenancy tests run export jobs. Running in parallel sometimes hits the per user limit of concurrent export requests, so we run them serially.

integration-tests/rbac-permission.test.ts
```diff
@@ -14,16 +14,16 @@ test('practitioner role can create new patient', async () => {
...
             response: { status: 401 },
         });
     });
 
     test('auditor role cannot create new patient record', async () => {
-        const client = await getFhirClient('auditor');
+        const client = await getFhirClient({ role: 'auditor' });
         await expect(client.post('Patient', randomPatient())).rejects.toMatchObject({
-            response: { status: 403 },
+            response: { status: 401 },
```
carvantes, line 26: This is needed because of https://github.com/awslabs/fhir-works-on-aws-routing/pull/92

That routing change was released but it has not been picked up by the mainline deployment package yet.

integration-tests/utils.ts
```diff
@@ -7,21 +7,76 @@ import * as AWS from 'aws-sdk';
...
+        USERNAME: username,
+        PASSWORD: password,
+    };
+};
+
+export const getFhirClient = async ({
+    role = 'practitioner',
+    providedAccessToken,
+    tenant = 'tenant1',
+}: { role?: 'auditor' | 'practitioner'; providedAccessToken?: string; tenant?: string } = {}) => {
```
rsmayda, line 78: This is returning an `axios` instance not this custom type right?

integration-tests/multitenancy.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expectResourceToNotBePartOfSearchResults,
+    getFhirClient,
+    randomPatient,
+    waitForResourceToBeSearchable,
+} from './utils';
+import BulkExportTestHelper from './bulkExportTestHelper';
+
+jest.setTimeout(300_000);
+
+test('empty test placeholder', () => {
```
rsmayda, line 17: Thoughts on just using `--passWithNoTests`? https://jestjs.io/docs/cli#--passwithnotests

integration-tests/multitenancy.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    // empty test to avoid the "Your test suite must contain at least one test." error
+});
+
+if (process.env.MULTI_TENANCY_ENABLED === 'true') {
+    describe('tenant data isolation', () => {
+        let client: AxiosInstance;
+        let clientForAnotherTenant: AxiosInstance;
+        beforeAll(async () => {
+            client = await getFhirClient({ tenant: 'tenant1' });
+            clientForAnotherTenant = await getFhirClient({ tenant: 'tenant2' });
```
rsmayda, line 27: In terms of set-up we created two Cognito users one with tenant1 and one with tenant2?

integration-tests/multitenancy.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import BulkExportTestHelper from './bulkExportTestHelper';
+
+jest.setTimeout(300_000);
+
+test('empty test placeholder', () => {
+    // empty test to avoid the "Your test suite must contain at least one test." error
+});
+
+if (process.env.MULTI_TENANCY_ENABLED === 'true') {
+    describe('tenant data isolation', () => {
```
rsmayda, line 22: An interesting test case I would like to try out is:
tenant1: has Patient/1234
tenant2: has Observation/1234 with subject set to `<url>/tenant1/Patient/1234`

Then do searches to see if we can break the isolation (I am thinking includes & rev_includes)

It may be easy to use two obscure resourceTypes to avoid search results getting too big

rsmayda (changes requested):
Could we add include/rev_include test cases?

integration-tests/multitenancy.test.ts
carvantes: It won't work. `--passWithNoTests` means allow passing without any test suite(i.e. there are no `*.test.ts` files). I don't think that there's a jest option to allow test suites with no tests.

integration-tests/multitenancy.test.ts
carvantes: yes 2 tenants are sufficient for our test cases. 

The default user for most tests (`COGNITO_USERNAME_PRACTITIONER`) is from `tenant1`. A new user (`COGNITO_USERNAME_PRACTITIONER_ANOTHER_TENANT`) is from `tenant2`

integration-tests/multitenancy.test.ts
```diff
@@ -63,6 +63,59 @@ if (process.env.MULTI_TENANCY_ENABLED === 'true') {
...
+                        reference: `${process.env.API_URL}/tenant/tenant1/Organization/${testOrganizationResource.id}`,
+                    },
+                })
+            ).data;
+
+            await waitForResourceToBeSearchable(clientForAnotherTenant, testPatientWithAbsoluteReferenceToOrg);
+
+            await expectResourceToNotBePartOfSearchResults(
+                clientForAnotherTenant,
+                { url: 'Patient', params: { _id: testPatientWithRelativeReferenceToOrg.id, _include: '*' } },
```
rsmayda, line 96: I don't think it would be an issue since we are passing in `_id`, but is it possible for these include/revincludes reaching page limits? i.e. the actual result is on a future page?

integration-tests/multitenancy.test.ts
carvantes: that's a good one. I added/revinclude tests with references to other tenant's resources.

integration-tests/multitenancy.test.ts
carvantes: That cannot happen. 

We are creating a new Patient specifically for this test. The search response will have exactly one resource(The Patient with the provided `_id`, and no other resources since the include/revinclude * should not match anything)

integration-tests/utils.ts
carvantes: That's not the return type of the function. That's the type of the parameters(It gets a bit tricky due to the defaults and destructuring of params). The return type of the function is implicitly `Promise<AxiosInstance>`

```ts
export const getFhirClient = async ({
    role = 'practitioner',
    providedAccessToken,
    tenant = 'tenant1',
}: { role?: 'auditor' | 'practitioner'; providedAccessToken?: string; tenant?: string } = {}) => {
...
```

It could be specified as:
```ts
export const getFhirClient = async ({
    role = 'practitioner',
    providedAccessToken,
    tenant = 'tenant1',
}: { role?: 'auditor' | 'practitioner'; providedAccessToken?: string; tenant?: string } = {}): Promise<AxiosInstance> => {
...
```

rsmayda: Ahh sorry yep that is right!

rsmayda (approved):

PR #393, opened 2021-08-02
test: Group export

Bingjiling:
Issue #, if available:

Description of changes:

* Integ test for group export

Checklist:


integration-tests/bulkExportTestHelper.ts
```diff
@@ -97,6 +116,36 @@ export default class BulkExportTestHelper {
...
+            // Update group members to resources just created
+            createGroupBundle.entry[1].resource.member = response.data.entry.map(
+                (item: { response: { location: any } }) => ({
+                    entity: { reference: item.response.location },
+                    ...groupMemberMeta,
+                }),
+            );
+            // Update reference in Encounter to reference a patient just created
+            // @ts-ignore
+            createGroupBundle.entry[0].resource.subject.reference = response.data.entry[0].response.location;
```
carvantes: Why use @ts-ignore here? 
That's a last-resort-kind-of-thing that can often be avoided

integration-tests/createGroupMembersBundle.json
```diff
@@ -0,0 +1,269 @@
...
+  "type": "transaction",
+  "entry": [
+    {
+      "fullUrl": "urn:uuid:fcfe413c-c62d-4097-9e31-02ff6ff523ad",
+      "resource": {
+        "resourceType": "Patient",
+        "id": "fcfe413c-c62d-4097-9e31-02ff6ff523ad",
+        "text": {
+          "status": "generated",
+          "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Generated by <a href=\"https://github.com/synthetichealth/synthea\">Synthea</a>.Version identifier: master-branch-latest-2-g5921215\n .   Person seed: 8914530193943894565  Population seed: 265</div>"
```
carvantes: Looks like Synthea was used to generate the test patients. That works, but I think that tests would be much easier to understand if we used patients/resources with only the minimum fields required for the test.

integration-tests/bulkExportTestHelper.ts
```diff
@@ -112,9 +161,12 @@ export default class BulkExportTestHelper {
...
     }
 
-    getResources(bundleResponse: any): Record<string, any> {
+    getResources(bundleResponse: any, originalBundle?: any): Record<string, any> {
         const resources = [];
-        const clonedCreatedBundle = cloneDeep(createBundle);
+        let clonedCreatedBundle = cloneDeep(createBundle);
+        if (originalBundle) {
+            clonedCreatedBundle = cloneDeep(originalBundle);
+        }
```
carvantes: I think we should make `createBundle` the default value for the `originalBundle` param (on the function params themselves, not with an if on the function body)

integration-tests/createGroupMembersBundle.json
```diff
@@ -0,0 +1,269 @@
...
+      },
+      "request": {
+        "method": "POST",
+        "url": "Practitioner"
+      }
+    },
+    {
+      "fullUrl": "urn:uuid:fcfe413c-c62d-4097-9e31-02ff6ff545bc",
+      "resource": {
+        "resourceType": "Group",
```
carvantes, line 54: Is this needed? I may be missing something, but I think that only one the other bundle(`createGroupAndPatientCompartmentBundle.json`) needs to declare a Group

integration-tests/bulkExportTestHelper.ts
```diff
@@ -97,6 +116,36 @@ export default class BulkExportTestHelper {
         }
     }
 
+    async createGroupAndReturnResources(groupMemberMeta?: GroupMemberMeta) {
+        try {
+            const response = await this.fhirUserAxios.post('/', createGroupMembersBundle);
+            const createGroupBundle = cloneDeep(createGroupAndPatientCompartmentBundle);
+            // Update group members to resources just created
+            createGroupBundle.entry[1].resource.member = response.data.entry.map(
```
carvantes: It's hard to tell what `createGroupBundle.entry[1]`, consider using `Array.find by resourceType == Group`

integration-tests/bulkExportTestHelper.ts
```diff
@@ -97,6 +116,36 @@ export default class BulkExportTestHelper {
...
+                (item: { response: { location: any } }) => ({
+                    entity: { reference: item.response.location },
+                    ...groupMemberMeta,
+                }),
+            );
+            // Update reference in Encounter to reference a patient just created
```
carvantes: I think that the bundle manipulation done to set up the test data can be greatly simplified.

Looks like all the data (with proper reference set between resources) for your test scenarios can be created in a single bundle transaction if you use internal references within the bundle. 

See: 
- https://www.hl7.org/fhir/bundle.html#references
- https://www.hl7.org/fhir/bundle-references.json.html

On the example json, the important part is that there's an `Observation` with:
```
"subject": {
          "reference": "urn:uuid:04121321-4af5-424c-a0e1-ed3aab1c349d"
        }
```

and a `Patient` with:
```
{
      "fullUrl": "urn:uuid:04121321-4af5-424c-a0e1-ed3aab1c349d",
      "resource": {
        "resourceType": "Patient",
```

---
An alternative would be to just create resources one at a time without using bundles. There's already a `randomPatient` method, and building a `Group` resource is easy. Although you probably want to do the Bundle thing since your code is already invested in using Bundles.

integration-tests/bulkExport.test.ts
```diff
@@ -70,4 +70,36 @@ describe('Bulk Export', () => {
...
+        const responseBody = await bulkExportTestHelper.getExportStatus(statusPollUrl);
+
+        // CHECK
+        return bulkExportTestHelper.checkResourceInExportedFiles(responseBody.output, {
+            ...members,
+            ...compartment,
+        });
+    });
+
+    test('Does not export inactive group members', async () => {
```
carvantes: Should we add a similar tests for members outside the membership period?

carvantes (changes requested):
Overall the tests appear to be correct. But the test setup code is a bit convoluted and can be simplified if we take advantage of advanced features of Bundle transactions (see my other comments)

integration-tests/createGroupMembersBundle.json
Bingjiling: Removed a group as a member in the group to make the test simpler

integration-tests/createGroupMembersBundle.json
Bingjiling: It's a hustle to update the fields for all resource types and not much value add. I'll leave this one as is.

integration-tests/bulkExportTestHelper.ts
```diff
@@ -62,7 +80,7 @@ export default class BulkExportTestHelper {
     }
 
     async getExportStatus(statusPollUrl: string, expectedSubstring = ''): Promise<any> {
-        const threeMinuteFromNow = new Date(new Date().getTime() + this.THREE_MINUTES_IN_MS);
+        const threeMinuteFromNow = new Date(new Date().getTime() + this.FIVE_MINUTES_IN_MS);
```
carvantes: rename the `threeMinuteFromNow` variable

integration-tests/createGroupMembersBundle.json
carvantes: I think It's not that hard to delete the extra fields, there are only 4 resources on the bundle.

The value is the clarity of the tests. The bundle has about 30 lines of actual relevant information, but any SDE needs to find them among 300+ lines of irrelevant stuff in order to understand the test data that is being set up.

carvantes (changes requested):

carvantes (approved):

PR #399, opened 2021-08-10
docs: update postman collection and docs to use Id token 

carvantes:
Description of changes:

Multi-tenancy requires the use of the Cognito `IdToken` instead of the `AccessToken` since only the `IdToken` can have custom claims to store the `tenantId`.

We updated the default single-tenant deployment to also use the `IdToken` since maintaining support for both token types adds complexity.

The postman collection was reverted to using the `COGNITO_AUTH_TOKEN` (which means running a script and copying the token into postman). It was possible to signin to Cognito from postman and get an access token, but postman makes it impossible to use the Id Token that way. See: https://github.com/postmanlabs/postman-app-support/issues/492, https://github.com/postmanlabs/postman-app-support/issues/8231

### DIFF
--- a/scripts/init-auth.py
+++ b/scripts/init-auth.py
@@ -22,4 +22,5 @@ response = client.initiate_auth(
     ClientId=sys.argv[1]
 )
 
-print(json.dumps(response['AuthenticationResult'], indent=2))
+id_token = response['AuthenticationResult']['IdToken']
+print(id_token)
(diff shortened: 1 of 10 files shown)


scripts/init-auth.py
```diff
@@ -22,4 +22,5 @@
     ClientId=sys.argv[1]
 )
 
-print(json.dumps(response['AuthenticationResult'], indent=2))
+id_token = response['AuthenticationResult']['IdToken']
+print(id_token)
```
Bingjiling, line 26: Apply the same change to `provision-user.py`?

Bingjiling (commented):
Looks good, a few other places might need change as well: 

GitHub workflow: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/feat-multitenancy/.github/workflows/deploy.yaml#L136 

Install script: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/feat-multitenancy/scripts/install.sh#L318 

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/feat-multitenancy/scripts/win_install.ps1#L296

scripts/init-auth.py
carvantes: Done. Also updated GH workflow and install scripts per your other comment

Bingjiling (approved):

PR #417, opened 2021-08-18
test: integ test for binary operations

Bingjiling:
Issue #, if available:

Description of changes:

Add integ tests for binary operations 

Checklist:

### DIFF
--- a/integration-tests/binary.test.ts
+++ b/integration-tests/binary.test.ts
@@ -1,0 +1,66 @@
+import axios, { AxiosInstance } from 'axios';
+import { getFhirClient } from './utils';
+
+jest.setTimeout(60 * 1000);
+
+describe('Binary operations', () => {
+    let client: AxiosInstance;
+    beforeAll(async () => {
+        client = await getFhirClient();
+    });
+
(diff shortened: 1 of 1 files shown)


integration-tests/binary.test.ts
```diff
@@ -0,0 +1,66 @@
...
+describe('Binary operations', () => {
+    let client: AxiosInstance;
+    beforeAll(async () => {
+        client = await getFhirClient();
+    });
+
+    const testFileContent = 'This is a test file for binary functionality';
+    const binaryTxtRequestBody = {
+        resourceType: 'Binary',
+        contentType: 'text/plain',
```
rsmayda, line 15: does file type matter in the integ test? I don't think so but want to verify

rsmayda (approved):

integration-tests/binary.test.ts
Bingjiling: It does not affect the data of the file. If a text file is uploaded using content type image/jpeg, the data will still be saved correctly, but the downloaded file will have extension `.jpeg` and can't be opened successfully till it's updated to `.txt`.

PR #419, opened 2021-08-18
chore(release): 4.0.0

carvantes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [4.0.0](https://github.com/awslabs/fhir-works-on-aws-deployment/compare/v3.1.0...v4.0.0) (2021-08-18)
+
+
+### ⚠ BREAKING CHANGES
+
+* The Cognito IdToken is now used instead of the accessToken to authorize requests.
+
+### Features
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+
+### ⚠ BREAKING CHANGES
+
+* The Cognito `IdToken` is now used instead of the `AccessToken` to authorize requests.
+
+Multi-tenancy itself is not a breaking change, you can continue to use FHIR works on single-tenant mode 
+by not using the `enableMultiTenancy` flag. 
+  
+However, note that updating an existing (single-tenant) stack to enable multi-tenancy is a breaking change. Multi-tenant 
+deployments use a different data partitioning strategy that renders the old, single-tenant, data inaccessible.
```
rsmayda, line 16: Do we have a script for this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fhir-works-on-aws-deployment",
-  "version": "3.1.0",
+  "version": "4.0.0",
```
Bingjiling, line 3: Need update to user agent string https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L57 as well

ssvegaraju (approved):

CHANGELOG.md
carvantes: There's no script to help make this migration.

Not every existing deployment out there benefits from switching to multi-tenancy. Some customers don't have multi-tenancy requirements and others prefer multiple separate (single-tenant) stacks for different tenants.

Migrating an existing stack from single-tenant to multi-tenant can be done by exporting all data (possibly with a single bulk export operation) and then inserting it into a new multi-tenant stack.

rsmayda: Let's add this task to back log I can see this being a feature some customers would like/need

package.json
carvantes: Done

rsmayda (approved):

PR #431, opened 2021-09-02
[WIP] feat: add config value for allow user scope for system export

Bingjiling:
Issue #, if available:

Description of changes:

* add config value for allow user scope for system export
* need to deploy with smart package to verify changes

Checklist:

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -17,6 +17,7 @@ custom:
   patientPickerEndpoint: ${opt:patientPickerEndpoint}
   useHapiValidator: ${opt:useHapiValidator, 'false'}
   enableMultiTenancy: ${opt:enableMultiTenancy, 'false'}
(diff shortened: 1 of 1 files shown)


serverless.yaml
```diff
@@ -110,6 +111,7 @@ functions:
     provisionedConcurrency: 5
     environment:
       EXPORT_STATE_MACHINE_ARN: !Ref BulkExportStateMachine
+      ALLOW_USER_SCOPE_FOR_SYSTEM_EXPORT: !Ref AllowUserScopeForSystemExport
```
rsmayda, line 114: I don't think this parameter will be used so another option is this to be configured via the config.ts file. May want to solicit opinions around the team.

Bingjiling: I like the idea of it to be configured via config.js file. Will close out this PR.

PR #437, opened 2021-09-08
fix: group export with group created before _since

Bingjiling:
Issue #, if available:

Description of changes:

* Before this fix, if a group was last updated before the _since timestamp in group export, the group will be removed form the dataframe before we reach the group filter step. This would cause the script to error out for index out of bound. 
* In this fix, the order of _since timestamp filter and group filter is shifted so group export will succeed with the members updated after the _since timestamp. 

Checklist:


rsmayda (approved):
LGTM

PR #440, opened 2021-09-14
chore: update nodejs runtime version to 14

ssvegaraju:
Issue #, if available:

Description of changes:
Updated `serverless.yaml` to use node js run time `nodejs14.x` instead. Deployed to AWS Account with no downtime or alarms in around 3-5 minutes.

Checklist:

### DIFF
--- a/.github/workflows/build-and-validate.yaml
+++ b/.github/workflows/build-and-validate.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: 14
       - name: Install dependencies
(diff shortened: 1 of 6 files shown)


rsmayda (changes requested):
Could we also update the node version in the package.json: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/package.json#L119

&& 

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/.github/workflows/deploy.yaml#L183

rsmayda (commented):
Currently waiting for our pipeline to clear prior to merging this one in

rsmayda (approved):

Issue #432, opened 2021-09-02
[Feature Request] Make config.ts async to allow for things like kms decrypt calls for ENV variables


medhost-jshaver:
**Is your feature request related to a problem? Please describe.**
Currently cannot make async calls from config.ts. I would like to pass kms encrypted ENV variables for the client_id/secret attributes of introspectionOptions and be able to call kms.decrypt() within config.ts/authZConfig.ts (for smart).

**Describe the solution you'd like**
I would be nice to have config.ts and authZConfig.ts (for smart) handle async methods.

**Describe alternatives you've considered**
We could make the kms decrypt call from the authz-smart package right before using the client_id/secret values for basic credentials, but this would be called for every request and seems like unnecessary overhead.

rsmayda:
Hey Josh,

Lets set up some time to discuss more!

carvantes:
This was added on https://github.com/awslabs/fhir-works-on-aws-deployment/pull/464

PR #443, opened 2021-09-16
feat: enable KMS decryption for token introspection secrets

medhost-jshaver:
Issue #, if available: https://github.com/awslabs/fhir-works-on-aws-deployment/issues/432

Description of changes:
made creation of fhirConfig async so I could use kms.decrypt for kms encrypted client_id & secret

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -19,6 +19,7 @@ custom:
   enableMultiTenancy: ${opt:enableMultiTenancy, 'false'}
   logLevel: ${opt:logLevel, 'error'}
   enableESHardDelete: ${opt:enableESHardDelete, 'false'}
+  enableTokenIntrospection: ${opt:enableTokenIntrospection, 'false'}
   patientCompartmentFileV3: 'patientCompartmentSearchParams.3.0.2.json'
   patientCompartmentFileV4: 'patientCompartmentSearchParams.4.0.1.json'
   bundle:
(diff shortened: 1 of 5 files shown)


rsmayda (commented):
Thanks for sharing this Josh let us review and discuss if we want to merge this in

carvantes:
Hi @medhost-jshaver ,
We like the idea of the added flexibility of async configuration. Please take a look at https://github.com/awslabs/fhir-works-on-aws-deployment/pull/465. That PR is based on this PR and it has the changes to make the creation of FhirConfig async.

The only notable difference is on `index.ts`, where special error handling was added. The thing is that provisioned concurrency in Nodejs does not wait for async initialization to finish. This could lead to "caching" a rejected promise on the `serverlessHandler` variable (i.e. if the call to KMS faces a transient network error and fails) and that'd make several Lambda requests fail without retrying the init code.

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/99e12283a24af1b4c6a3b4f709b5618d6b656e1b/src/index.ts#L10-L19

Unfortunately we won't be merging this PR at this moment as the KMS decryption and the env variables that you use seem very specific to your use case. We suggest to rebase your changes on top of https://github.com/awslabs/fhir-works-on-aws-deployment/pull/465

medhost-jshaver:
Thanks a ton for adding this feature and for fixing that issue with the potentially bad promise cache @carvantes! I totally didn't think about that scenario.

PR #448, opened 2021-09-20
chore: resolve transitive dependency on set-value package

ssvegaraju:
Issue #, if available:

Description of changes:
Small update to resolutions to address security vulnerabilities in set-value versions <4.0.1. Deployed to AWS Account successfully in ~3 minutes, and passed integration tests and local tests.

Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,8 @@
     "ws": "^7.4.6",
     "trim-newlines": "^3.0.1",
     "normalize-url": "^4.5.1",
-    "path-parse": "^1.0.7"
+    "path-parse": "^1.0.7",
+    "set-value": "^4.0.1"
   },
   "jest": {
     "moduleFileExtensions": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -89,7 +89,8 @@
     "ws": "^7.4.6",
     "trim-newlines": "^3.0.1",
     "normalize-url": "^4.5.1",
-    "path-parse": "^1.0.7"
+    "path-parse": "^1.0.7",
+    "set-value": "^4.0.1"
```
rsmayda, line 93: should this update our `yarn.lock` too? did you run `yarn install`?

ssvegaraju: It absolutely should, thanks for catching that! Pushed an update to make sure that that change is present.

carvantes (approved):
lgtm. Also, I updated the title to be `chore:` instead of `fix:`. Dependency updates are always chore or similar. We don't want them to appear on changelog.

rsmayda (approved):

PR #452, opened 2021-09-21
fix: handle when failures happen in bulk export

rsmayda:
Description of changes:
- Previously `catchAllUpdateStatusToFailed` step in our step function did not pass the required `jobId` to update the DDB table. This resulted in the all future exports being broken due to the system believing the export is currently running.
- **To fix:** we need to put the error in a different json path & have this step function have the same input as the other status update ones.
  -  https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html#input-output-resultpath-catch
- Most of the file is yaml linting to match our other CFN files. I will highlight the 2 material changes

Testing:
- I tested this by focusing the `getJobStatus` step to throw an error and ensuring that the `catchAllUpdateStatusToFailed` step ran correctly and updated the DDB

Checklist:

### DIFF
--- a/bulkExport/state-machine-definition.yaml
+++ b/bulkExport/state-machine-definition.yaml
@@ -5,51 +5,52 @@
 
 id: BulkExportStateMachine
 definition:
-  Comment: "State machine that executes a FHIR bulk export job"
+  Comment: 'State machine that executes a FHIR bulk export job'
   StartAt: parallelHelper
   States:
     catchAllUpdateStatusToFailed:
       Type: Task
       Resource: !GetAtt updateStatus.Arn
-      Parameters: {"jobId.$":"$.jobId", "status": "failed"}
+      Parameters: { 'globalParams.$': '$', 'status': 'failed' }
       Retry:
-        - ErrorEquals: [ "States.ALL" ]
+        - ErrorEquals: ['States.ALL']
       End: true
     parallelHelper:
       Type: Parallel
       End: true
       Catch:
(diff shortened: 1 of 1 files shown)


bulkExport/state-machine-definition.yaml
```diff
@@ -5,51 +5,52 @@
...
 definition:
-  Comment: "State machine that executes a FHIR bulk export job"
+  Comment: 'State machine that executes a FHIR bulk export job'
   StartAt: parallelHelper
   States:
     catchAllUpdateStatusToFailed:
       Type: Task
       Resource: !GetAtt updateStatus.Arn
-      Parameters: {"jobId.$":"$.jobId", "status": "failed"}
+      Parameters: { 'globalParams.$': '$', 'status': 'failed' }
```
rsmayda, line 14: Material change here

bulkExport/state-machine-definition.yaml
```diff
@@ -5,51 +5,52 @@
...
+        - ErrorEquals: ['States.ALL']
       End: true
     parallelHelper:
       Type: Parallel
       End: true
       Catch:
-        - ErrorEquals: [ "States.ALL" ]
+        - ErrorEquals: ['States.ALL']
           Next: catchAllUpdateStatusToFailed
+          ResultPath: '$.error'
```
rsmayda, line 24: Material change here

ssvegaraju (approved):

PR #466, opened 2021-09-30
fix: remove transient search crucible test

ssvegaraju:
Issue #, if available:

Description of changes:
pointed crucible tests to new branch, where search test that resulted in transient error was removed. I deployed to my AWS account and ensured that all tests passed as expected.

Checklist:

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -135,7 +135,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: nguyen102/plan_executor
-          ref: r4-aws-fhir-solution
+          ref: r4-aws-fhir-fixed-search
       - uses: actions/setup-ruby@v1
         with:
           ruby-version: '2.6'


.github/workflows/deploy.yaml
```diff
@@ -135,7 +135,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: nguyen102/plan_executor
-          ref: r4-aws-fhir-solution
+          ref: r4-aws-fhir-fixed-search
```
carvantes, line 138: I think that we should merge the fix into the `r4-aws-fhir-solution` of `nguyen102/plan_executor` and not modify this file

carvantes (changes requested):

.github/workflows/deploy.yaml
ssvegaraju: Gotcha, I'll make a PR to merge changes into the `r4-aws-fhir-solution` branch instead, and close this PR.

PR #467, opened 2021-09-30
fix: Fix Implentation guide integration test

nguyen102:
Issue #, if available:

Description of changes:
In the IG test we're making a request to https://www.hl7.org/fhir/us/core/STU3.1.1/CapabilityStatement-us-core-server.json. We're comparing our Cap Statement with the official FHIR Cap statement. Unfortunately the SSL cert at that domain has expired. The fix in this PR allows us to make a request to that URL even though the SSL cert on that domain has expired.

Checklist:

### DIFF
--- a/integration-tests/implementationGuides.test.ts
+++ b/integration-tests/implementationGuides.test.ts
@@ -3,6 +3,7 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 import axios, { AxiosInstance } from 'axios';
+import https from 'https';
 import waitForExpect from 'wait-for-expect';
 import { cloneDeep } from 'lodash';
 import { Chance } from 'chance';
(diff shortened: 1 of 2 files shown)


ssvegaraju (approved):

ssvegaraju (approved):
Can the security warning be dismissed?

integration-tests/implementationGuides.test.ts
```diff
@@ -94,7 +95,14 @@ describe('Implementation Guides - US Core', () => {
         );
 
         const expectedCapStatement: CapabilityStatement = (
-            await axios.get(`https://www.hl7.org/fhir/us/core/${usCoreVersion}/CapabilityStatement-us-core-server.json`)
+            await axios.get(
+                `https://www.hl7.org/fhir/us/core/${usCoreVersion}/CapabilityStatement-us-core-server.json`,
+                {
+                    httpsAgent: new https.Agent({
+                        rejectUnauthorized: false,
```
carvantes: It's probably better to download the capability statement from https://www.hl7.org/fhir/us/core/STU3.1.1/CapabilityStatement-us-core-server.json and commit the file as part of the integ tests code. That gets rid of any other transient issue (i.e. www.hl7.org is down) and the json files are not that big.

Plus, disabling certificate validation is an odd choice that will be flagged by any security scan. (CodeQL already flagged it). We could dismiss it as "used in tests" but that's less than ideal

carvantes (changes requested):

integration-tests/implementationGuides.test.ts
nguyen102: That makes sense. I just updated the code to use the downloaded JSON.

carvantes (approved):

PR #474, opened 2021-10-07
feat: add custom resource to update search mappings

carvantes:
Add a custom resource to update the search mappings as part of the CFN deployment.

There are 3 scenarios, **the custom resource is expected to always succeed**:
1. A brand new stack is created.
    Custom resource creates Indices for all resource types with the static mappings
2. An existing stack is updated and the existing indices are "compatible" with the static mappings. ("compatible" here means that there is no conflict of types for any field. This happens when existing dynamic mappings were created with perfectly valid FHIR resources)
  Custom resource updates the mappings for existing indices (this may have no effect since dynamic mappings are a superset of the static mappings) and creates any missing index so that all resource types have an index.
3. An existing stack is updated and the existing indices are "incompatible" with the static mappings. (This happens when the dynamic mappings were created with invalid FHIR resources and some field types are incorrect or maybe if the customer manually modified the mappings)
  Custom resource creates missing indexes and attempts to update the mappings for existing indexes. If there are errors when updating the mappings (due to "incompatible mappings"), the error is logged as a warning and the execution continues. The custom resource succeeds.

**Note:** The build fails because we need to first release the change on the search package https://github.com/awslabs/fhir-works-on-aws-search-es/pull/126

Checklist:


Bingjiling (approved):
LGTM

serverless.yaml
```diff
@@ -16,6 +16,7 @@ custom:
   config: ${file(serverless_config.json)}
   useHapiValidator: ${opt:useHapiValidator, 'false'}
   enableMultiTenancy: ${opt:enableMultiTenancy, 'false'}
+  skipSearchMappingsUpdate: ${opt:skipSearchMappingsUpdate, 'false'}
```
rsmayda: So by default we are turning on the mappings;; could this break any customer?

rsmayda: Is it more understandable to say `updateSearchMappings` instead of `skip...` as a paramName?

cloudformation/elasticsearch.yaml
```diff
@@ -144,3 +144,10 @@ Resources:
               Resource:
                 Fn::Sub: arn:${AWS::Partition}:es:${AWS::Region}:${AWS::AccountId}:domain/*
         - !Ref AWS::NoValue
+  UpdateSearchMappingsCustomResource:
+    DependsOn: ElasticSearchDomain
+    Condition: doSearchMappingsUpdate
+    Type: AWS::CloudFormation::CustomResource
+    Properties:
+      ServiceToken: !GetAtt UpdateSearchMappingsLambdaFunction.Arn # serverless by convention capitalizes first letter and suffixes with "LambdaFunction"
```
rsmayda, line 153: Unrelated to this line

Do we need to add more restrictive validation to our persistence layer? With "stricter" indices writes to the persistence layer would succeed and maybe fail more often on the search layer

carvantes: I don't think this changes the way we treat validation. It's true that poor validation (In theory we even allow disabling validation) may result in ddbToEs dlq messages since the resources do not match the ES mappings, but I don't think that'll happen more often

In a sense, the static mappings are actually less strict than the dynamic mappings, since static mappings only index the subset of the fields that are used in search params, whereas dynamic mappings index ALL fields.

rsmayda: That second part is something I didn't consider. Sounds good thanks for explaining

serverless.yaml
carvantes: I ended up removing the `skipSearchMappingsUpdate` flag altogether. It was indeed a (small) breaking change. Existing customers could have seen errors on the CFN deployment and would have needed to add the `skipSearchMappingsUpdate` flag.

I ended up modifying the behavior on the search package (https://github.com/awslabs/fhir-works-on-aws-search-es/pull/126) so that it can be configured to ignore errors for the update mappings requests (the `ignoreMappingsErrorsForExistingIndices` param). We'll start off with this param set to true.

This way, all new customers get the new static mappings and it is a non-breaking change for existing customers. Existing customers benefit from this change since any missing index is created with the static mappings, they can check the logs for warnings about mapping conflicts on existing indices and manually reindex with the new mappings, or they ignore them and continue using the already created dynamic mappings (In many cases they work just fine).

rsmayda: To confirm will our customers see the warnings about mapping conflicts when they run `sls deploy`? or where would they see the warnings

carvantes: warnings can be seen in the logs of the `updateSearchMappings` Lambda function.

rsmayda: Does `disableLogs: true # needed to avoid race condition error "Resource of type 'AWS::Logs::LogGroup' already exists" since the custom resource lambda invocation may create the log group before CFN does` in the CFN stop that from happening

cloudformation/elasticsearch.yaml
```diff
@@ -144,3 +144,9 @@ Resources:
               Resource:
                 Fn::Sub: arn:${AWS::Partition}:es:${AWS::Region}:${AWS::AccountId}:domain/*
         - !Ref AWS::NoValue
+  UpdateSearchMappingsCustomResource:
```
rsmayda, line 149: Could you add a comment to line 96 saying something along the lines of "if you need to change your EBS or cluster configurations please also consider changing your shard count `NUMBER_OF_SHARDS`"

serverless.yaml
carvantes: In general there's no way to print progress/warning logs on the console when running `sls deploy`.  The exception is when the CFN deployment fails, in that case `sls deploy` prints the failure message.

The `disableLogs: true` is needed for a very peculiar issue with custom resources. Without it the deployment sometimes fails due to "log group with same name already exists error" because serverless attempts to create a log group, but the log group was created seconds before the moment the custom resource started running (the 1st lambda invocation naturally creates a log group )

cloudformation/elasticsearch.yaml
carvantes: sure, I just added a comment there

rsmayda (approved):

rsmayda (approved):

PR #487, opened 2021-10-12
chore: update bulk tests to not allow user scopes

rsmayda:
Description of changes:
Due to [smart-authz dependency change](https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/CHANGELOG.md#300-2021-09-15) we no longer allow only `user` scopes to do the export. This commit updates our tests to reflect that.

Checklist:

### DIFF
--- a/integration-tests/bulkExport.test.ts
+++ b/integration-tests/bulkExport.test.ts
@@ -19,8 +19,10 @@ describe('Bulk Export', () => {
 
         beforeAll(async () => {
             const fhirUserAxios = await getFhirClient('fhirUser user/*.*', true);
-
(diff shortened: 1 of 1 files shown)


integration-tests/bulkExport.test.ts
```diff
@@ -72,6 +74,20 @@ describe('Bulk Export', () => {
             // CHECK
             return bulkExportTestHelper.getExportStatus(statusPollUrl, 'Export job has been canceled');
         });
+
+        test('Ensure just `user` scope cannot do a system export', async () => {
```
rsmayda:
```suggestion
        test('Ensure `user` scope cannot do a system export', async () => {
```

Bingjiling (approved):

Issue #390, opened 2021-07-28
[Bug] Unable to update Implementation Guide: Lambda too big to deploy with IG


ericfuxealth:
**Describe the bug**
Trying to follow the steps here to add implementation guide but ran into issues with the Lambda package too big to deploy.
https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/USING_IMPLEMENTATION_GUIDES.md

**To Reproduce**
1. Follow the steps here https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/USING_IMPLEMENTATION_GUIDES.md
2. Download the US core package from https://www.hl7.org/fhir/us/core/package.tgz
3. Unzip the package, rename the folder to "hl7.fhir.us.core"
4. cd hl7.fhir.us.core
5. Run "npm --registry https://packages.simplifier.net i" 
6. mv node_modules/* .. 
I now have these packages:
- hl7.fhir.r4.core	
- hl7.fhir.us.core	
- hl7.fhir.uv.bulkdata	
- us.nlm.vsac

7. cd .. (now in fhir-works-on-aws-deployment)
8. run "yarn run compile-igs"
9. cd javaHapiValidatorLambda (now  in fhir-works-on-aws-deployment/javaHapiValidatorLambda)
10. mvn clean install
11. serverless deploy
12. Error " An error occurred: ValidatorLambdaFunction - Resource handler returned message: "Unzipped size must be smaller than 262144000 bytes (Service: Lambda, Status Code: 400, Request ID: 8a035ca9-1d30-4993-baa0-8b837d71c684, Extended Request ID: null)" (RequestToken: 3704c919-1ac0-ca0b-5571-7cb26bda4b19, HandlerErrorCode: InvalidRequest)."

I checked the unzipped jar file size, the issue the implementation guides are too big.

**Expected behavior**
No error installing the Lambda

**Versions (please complete the following information):**
 - Release Version installed 3.0.0

**Additional context**
Are the above steps correct?

SanketD92:
Hello @ericfuxealth, thank you for reporting this error. We've added it to our backlog and we'll work it into our sprints based on priority.

Bingjiling:
[...]

PR #492, opened 2021-10-14
fix: use S3 bucket for implementation guides

ssvegaraju:
Issue #, if available:

Description of changes:
Set up the new framework for uploading/downloading implementation guides with S3 Buckets. There is room for improvement by implementing multithreading on S3 bucket download, but as it stands deployment takes approximately 10-15 minutes. Revised validator tests to use new S3 bucket method with a mocking library. Need to update documentation.

Checklist:


javaHapiValidatorLambda/pom.xml
```diff
@@ -174,23 +179,6 @@
     </dependencies>
 
     <build>
-        <resources>
-            <!-- Copy the implementation guides packages from resources folder -->
-            <resource>
```
rsmayda, line 179: Why do we want to get rid of these?

javaHapiValidatorLambda/serverless.yml
```diff
@@ -48,9 +58,70 @@ resources:
...
+                Version: '2012-10-17'
+                Statement:
+                  - Effect: Allow
+                    Action:
+                      - s3:PutObject
+                      - s3:GetObject
+                      - s3:ListObjectsV2
+                      - s3:ListBuckets
+                      - s3:DeleteObject
+                      - s3:*
```
rsmayda: We should remove this, prior to merging

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +39,18 @@ public Handler() {
...
+        if (region == null) {
+            region = "us-west-2";
+        }
+        final AmazonS3 s3 = AmazonS3ClientBuilder.standard().withRegion(Regions.fromName(region)).build();
+        String bucketName = getIGBucketName(s3.listBuckets());
+        if (bucketName.isEmpty()) {
+            throw new Error("Unable to find S3 bucket for implementation guides.");
+        }
+        List<String> objectKeys = getBucketObjects(s3, bucketName, region);
+        String guidesDirectory = downloadObjects(s3, bucketName, objectKeys);
```
rsmayda: Should we put this outside the handler so that these values are cached; Nestor may have a better insight on this

Bingjiling:
Looks like there's some lint issue?

javaHapiValidatorLambda/serverless.yml
```diff
@@ -23,17 +23,27 @@ provider:
...
+    role: IGBucketLambdaRole
+    runtime: nodejs14.x
+    description: 'Upload implementation guides to s3'
+    environment:
+      IMPLEMENTATION_GUIDES_BUCKET: !Ref FHIRImplementationGuidesBucket
   validator:
     handler: software.amazon.fwoa.Handler
+    role: IGBucketLambdaRole
     timeout: 25
     memorySize: 2048
```
rsmayda: Do we need to increase Lambda memory? or is 2GB enough. May want to add comment saying you may need to scale this to fit your IG size

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +65,50 @@ public ValidatorResponse handleRequest(String event, Context context) {
         return validate;
 
     }
+
+    private List<String> getBucketObjects(AmazonS3 s3, String bucketName, String regionString) {
+
+        ListObjectsV2Result result = s3.listObjectsV2(bucketName);
```
rsmayda: What is the S3 bucket structure? all IGs are in a single bucket or is it done by folders?

If done by folders you can use prefixes:
`new ListObjectsV2Request().withBucketName(bucket.getName()).withPrefix(prefix).withDelimiter(DELIMITER);`

https://stackoverflow.com/questions/38051789/listing-files-in-a-specific-folder-of-a-aws-s3-bucket

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +39,18 @@ public Handler() {
...
             fhirVersion = Validator.FHIR_R4;
         }
-
-        validator = new Validator(fhirVersion);
+        String region = System.getenv("REGION");
+        if (region == null) {
+            region = "us-west-2";
+        }
+        final AmazonS3 s3 = AmazonS3ClientBuilder.standard().withRegion(Regions.fromName(region)).build();
+        String bucketName = getIGBucketName(s3.listBuckets());
```
rsmayda: The bucket name could be passed in as an environment variable

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +65,50 @@ public ValidatorResponse handleRequest(String event, Context context) {
         return validate;
 
     }
+
+    private List<String> getBucketObjects(AmazonS3 s3, String bucketName, String regionString) {
+
+        ListObjectsV2Result result = s3.listObjectsV2(bucketName);
+        List<S3ObjectSummary> objects = result.getObjectSummaries();
```
rsmayda: You will need to take into account the `continuation-token`: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax

i.e. if there are so many items you need to iteratively get them

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +65,50 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+                    outputStream.write(buffer, 0, length);
+                }
+                inputStream.close();
+                outputStream.close();
+            }
+            return igLocation.toAbsolutePath().toString();
+        } catch (IOException e) {
+            throw new Error(e.getMessage());
+        } catch (AmazonServiceException e) {
+            throw new Error(e.getErrorMessage());
```
rsmayda: You don't need to recast the error into a more generic error
```suggestion
        } catch (IOException e) {
            throw e;
        } catch (AmazonServiceException e) {
            throw e;
```

javaHapiValidatorLambda/serverless.yml
```diff
@@ -23,17 +23,27 @@ provider:
...
-    provisionedConcurrency: 5
+#    provisionedConcurrency: 5
 
 resources:
```
rsmayda, line 51: It looks like you still need to create the Custom resource; see code example here: https://github.com/awslabs/fhir-works-on-aws-deployment/commit/e941aa74e38a24373b90e520f689e3e8d7c24689

Look for `UpdateSearchMappingsCustomResource`

javaHapiValidatorLambda/uploadIGToS3.ts
```diff
@@ -0,0 +1,92 @@
...
+            console.log('Deleting files from s3');
+            await Promise.all(
+                getFilesInDir('implementationGuides', 'json', []).map(entry => {
+                    console.log(`uploading ${entry.filename} to ${process.env.IMPLEMENTATION_GUIDES_BUCKET}`);
+                    return s3
+                        .deleteObject({
+                            Bucket: process.env.IMPLEMENTATION_GUIDES_BUCKET,
+                            Key: entry.filename,
+                        })
+                        .promise();
```
rsmayda: Do we actually want to delete the objects in the S3 bucket? Our default behavior is "no"

rsmayda (changes requested):

javaHapiValidatorLambda/pom.xml
ssvegaraju: this was the old way of attaching implementation guides through maven, hence us removing it here.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: I will look into caching these values so we don't have to construct an S3 client every time.

javaHapiValidatorLambda/.factorypath
```diff
@@ -0,0 +1,75 @@
+<factorypath>
```
carvantes: Remove this file. 

Looks like this is generated by eclipse and should not be committed. Try adding it to .gitignore as well

javaHapiValidatorLambda/serverless.yml
```diff
@@ -5,10 +5,15 @@
...
 
+plugins:
+  - serverless-plugin-scripts
 custom:
   stage: ${opt:stage, self:provider.stage}
   fhirVersion: ${opt:fhirVersion, '4.0.1'}
   region: ${opt:region, self:provider.region}
+  # scripts:
+  #   hooks:
+  #     'before:deploy:deploy': node uploadIGToS3.js ${self:custom.region}
```
carvantes: Why are we using serverless hooks instead of a CFN custom resource?

Also, looks like this is commented out by mistake.

javaHapiValidatorLambda/serverless.yml
```diff
@@ -23,7 +28,18 @@ provider:
...
-
+    REGION:
+      !Ref Region
+  iamRoleStatements:
+    - Effect: "Allow"
+      Action:
+        - "s3:listObjectsV2"
+        - "s3:getObject"
+        - "s3:invoke"
+        - "s3:*"
```
carvantes: Remove `"s3:*"`. 

Also, `"s3:invoke"` and `"s3:listObjectsV2"` do not exist. Sometimes there are higher level APIs that don't have a matching IAM action. Check the documentation to see what permissions are needed: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html (looks like listObjectsV2 needs `s3:ListBucket`)

javaHapiValidatorLambda/serverless.yml
```diff
@@ -48,9 +64,36 @@ resources:
           - '4.0.1'
           - '3.0.1'
         Description: 'The FHIR version used by the validator. Default: 4.0.1'
+      Region:
```
carvantes: I don't understand why this is needed. I think that we need to remove this and all references to the env variable.

CFN stacks are created in the region where `cfn:createStack` is executed, we don't need to add a parameter to tell us which region to use.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +45,14 @@ public Handler() {
         if (fhirVersion == null){
             fhirVersion = Validator.FHIR_R4;
         }
-
-        validator = new Validator(fhirVersion);
+        String region = System.getenv("REGION");
+        if (region == null) {
+            region = "us-west-2";
+        }
```
carvantes: why?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +45,14 @@ public Handler() {
...
+        final AmazonS3 s3 = AmazonS3ClientBuilder.standard().withCredentials(new EnvironmentVariableCredentialsProvider()).withRegion(Regions.fromName(region)).build();
```
carvantes: This works but you can simply do: `AmazonS3 s3Client = AmazonS3ClientBuilder.defaultClient();`

This will run in Lambda, so it's better to let the SDK pick up the credentials/region from the Lambda environment where it is running.

https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html

javaHapiValidatorLambda/uploadIGToS3.js
```diff
@@ -0,0 +1,102 @@
...
+        uploadAllFiles(() => {
+            console.log("successfully uploaded implementation guides to S3");
+            process.exit(0);
+        })
+    })
+    .catch(err => {
+        console.log("No bucket found, creating one...");
+        s3.createBucket(
+            {
+                Bucket: bucketName,
```
carvantes: s3 bucket should be defined in the CFN template, not created here. 
Also, hardcoding the s3 bucket name to `'fhir-service-validator-implementationguides'` causes all sorts of trouble, it's better to let CFN choose a unique bucket name

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -7,13 +7,36 @@
...
+import com.amazonaws.services.s3.model.S3ObjectSummary;
 
 import lombok.extern.slf4j.Slf4j;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 
 @Slf4j
 public class Handler implements RequestHandler<String, ValidatorResponse> {
 
     private final Validator validator;
+    private String bucketName = "fhir-service-validator-implementationguides";
```
carvantes: the bucket name should be passed as an env variable to the lambda function

carvantes (changes requested):

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +67,48 @@ public ValidatorResponse handleRequest(String event, Context context) {
         return validate;
 
     }
+
+    private String getIGBucketName(List<Bucket> buckets) {
```
carvantes: looks like this is unused

javaHapiValidatorLambda/uploadIGToS3.js
```diff
@@ -0,0 +1,102 @@
...
+                process.exit(1);
+                reject();
+            }
+            console.log(`File Uploaded successfully. ${data.Location}`);
+            resolve();
+        });
+    })
+}
+
+async function uploadAllFiles (callback) {
```
carvantes: I think that this needs to first delete the old files from S3.

e.g. If a customer is using US Core and CARIN BB, and then changes to using exclusively US Core. We need to somehow delete the CARIN BB files

carvantes (changes requested):

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +67,48 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+                FileOutputStream outputStream = new FileOutputStream(File.createTempFile("/tmp/" + keys.get(i).substring(0, keys.get(i).length() - 5), ".json"));
+                byte[] buffer = new byte[1024];
+                int length = 0;
+                while ((length = inputStream.read(buffer)) > 0) {
+                    outputStream.write(buffer, 0, length);
+                }
+                inputStream.close();
+                outputStream.close();
+            }
+            return "/tmp";
```
carvantes: We want downloads to be fast and occur in parallel if possible. We probably don't need to code it ourselves, you can use a higher level API class like [TransferManager](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/transfer/TransferManager.html) (I haven't used it, there may be a better option)

Although downloading to `/tmp` is not ideal since it has a 512MB limit. We could download to memory and load them directly into the validator instead.

Using `/tmp` may be ok for simplicity if it gives the desired results. 
How did you test it?
Is this enough to fix https://github.com/awslabs/fhir-works-on-aws-deployment/issues/390?
How big are those IG files?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +45,14 @@ public Handler() {
...
-
-        validator = new Validator(fhirVersion);
+        String region = System.getenv("REGION");
+        if (region == null) {
+            region = "us-west-2";
+        }
+        final AmazonS3 s3 = AmazonS3ClientBuilder.standard().withCredentials(new EnvironmentVariableCredentialsProvider()).withRegion(Regions.fromName(region)).build();
+        List<String> objectKeys = getBucketObjects(s3, region);
+        String guidesDirectory = downloadObjects(s3, objectKeys);
+        validator = new Validator(fhirVersion, guidesDirectory);
```
carvantes: Does this work? The validator uses `ClassGraph` to load the IGs (they used to be budled in the jar). Now that they are files in `/tmp` I don't think that they can be found

https://github.com/classgraph/classgraph

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: It works in the case of #390 as the total unzipped size of the files is 257MB, but I will try and shift to the CFN method you described above rather than using serverless hooks, and try to get the validator to accept that into the constructor instead of specifying a directory path.

javaHapiValidatorLambda/serverless.yml
ssvegaraju: Based on some research, it appears that 5G is the sweet spot between maintaining costs and increasing efficiency.

Bingjiling:
Great work! Could you please add a description on what steps a customer should take to deploy the validator with this change? And what's the sequence of things happening(e.g. S3 bucket create -> IG upload, etc)?

ssvegaraju:
> Great work! Could you please add a description on what steps a customer should take to deploy the validator with this change? And what's the sequence of things happening(e.g. S3 bucket create -> IG upload, etc)?

Sure, will add an update to the documentation shortly!

javaHapiValidatorLambda/createIGBucket/serverless.yaml
```diff
@@ -0,0 +1,113 @@
...
+              - Sid: "LambdaUploadAllowPolicy"
+                Effect: "Allow"
+                Principal:
+                  Service: 'lambda.amazonaws.com'
+                Action:
+                  - "s3:*"
+                Resource:
+                  - !GetAtt ImplementationGuidesBucket.Arn
+                  - !Join ['', [!GetAtt ImplementationGuidesBucket.Arn, '/*']]
+      UploadImplementationGuidesLambdaRole:
```
Bingjiling, line 84: Why do we need this role? Does the s3Sync plugin use Lambda to upload the files to S3 bucket? 

Also, we need to install the S3Sync plugin before using it here.

javaHapiValidatorLambda/createIGBucket/serverless.yaml
```diff
@@ -0,0 +1,113 @@
...
+                  - Effect: Allow
+                    Action:
+                      - "s3:PutObject"
+                      - "s3:GetObject"
+                      - "s3:CreateBucket"
+                      - "s3:ListBucket"
+                      - "s3:ListObjects"
+                    Resource:
+                      - !Join ['', ['arn:aws:s3:::', !Ref ImplementationGuidesBucket, '/*']]
+                      - !Join ['', ['arn:aws:s3:::', !Ref ImplementationGuidesBucket]]
```
Bingjiling, line 113: Add new line at end

javaHapiValidatorLambda/serverless.yml
```diff
@@ -23,16 +23,29 @@ provider:
   environment:
     FHIR_VERSION:
       !Ref FhirVersion
-
+    IMPLEMENTATION_GUIDES_BUCKET: !ImportValue ImplementationGuidesBucketNameKey
+  iam:
```
Bingjiling, line 27: Put the role on the function level? If we add more functions to this stack in the future, it'll be automatically applied to the new function and might create confusion.

javaHapiValidatorLambda/serverless.yml
```diff
@@ -23,16 +23,29 @@ provider:
...
 package:
   artifact: target/fwoa-hapi-validator-dev.jar
 
 functions:
   validator:
     handler: software.amazon.fwoa.Handler
-    timeout: 25
-    memorySize: 2048
+    timeout: 300
+    memorySize: 5120
```
Bingjiling, line 48: Add documentation in the implementation guide doc suggesting updating the memory size here if user are facing IG size issue?

javaHapiValidatorLambda/serverless.yml
```diff
@@ -53,4 +66,4 @@ resources:
         Description: Validator lambda function alias
         Value: !Ref ValidatorProvConcLambdaAlias # serverless by convention capitalizes first letter and suffixes with "ProvConcLambdaAlias"
         Export:
-          Name: !Join ['-', ['fhir-service-validator-lambda', !Ref Stage]]
+          Name: !Join ['-', ['fhir-service-validator-lambda', !Ref Stage]]
```
Bingjiling, line 69: Add new line

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -22,8 +35,13 @@ public Handler() {
         if (fhirVersion == null){
             fhirVersion = Validator.FHIR_R4;
         }
-
-        validator = new Validator(fhirVersion);
+        String bucketName = System.getenv("IMPLEMENTATION_GUIDES_BUCKET");
```
Bingjiling, line 38: Put this at class level?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +56,49 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+                req.setContinuationToken(token);
+            } while (result.isTruncated());
+            log.info("found " + objectKeys.size() + " keys");
+            return objectKeys;
+        } catch (Exception e) {
+            throw new Error(e);
+        }
+    }
+
+    private DownloadedGuidesHolder downloadObjects(List<String> keys, String bucketName) {
```
Bingjiling, line 83: It is a bit overkill to create a class for DownloadedGuidesHolder if it's just used to hold two list. It could return something like `Map<String, List<IGObject>>` with two keys `indices` and `resources`.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +56,49 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+        // keep track of .index.json objects
+        List<IGObject> indices = new ArrayList<IGObject>();
+        List<IGObject> resources = new ArrayList<IGObject>();
+        for (String key : keys) {
+            try (S3ObjectInputStream s3Object = s3.getObject(bucketName, key).getObjectContent()) {
+                IGObject bucketObj = new Utils.IGObject(key, IOUtils.toString(s3Object));
```
Bingjiling, line 89: Can we build this `toString` functionality into the `IGObject` class? So `IGObject` can have a `getContentObject` and `getContentString` method. This way it gives more flexibility to downstream logic to choose from what to use.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -64,22 +61,24 @@
     private final FhirContext ctx;
 
     private final String fhirVersion;
-    private final String implementationGuidesFolder;
+    private final List<IGObject> implementationGuidesIndices;
+    private final List<IGObject> implementationGuidesResources;
 
     public Validator() {
```
Bingjiling, line 67: If the `List<IGObject> DEFAULT_IMPLEMENTATION_GUIDES` is null now, would the first two constructors still be valid?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -64,22 +61,24 @@
...
-        this(FHIR_R4, DEFAULT_IMPLEMENTATION_GUIDES_FOLDER);
+        this(FHIR_R4, DEFAULT_IMPLEMENTATION_GUIDES, DEFAULT_IMPLEMENTATION_GUIDES);
     }
 
     public Validator(String fhirVersion) {
-        this(fhirVersion, DEFAULT_IMPLEMENTATION_GUIDES_FOLDER);
+        this(fhirVersion, DEFAULT_IMPLEMENTATION_GUIDES, DEFAULT_IMPLEMENTATION_GUIDES);
     }
 
-    public Validator(String fhirVersion, String implementationGuidesFolder) {
```
Bingjiling, line 77: Do we need a constructor `Validator(List<IGObject> indices, List<IGObject> definitions)` as well?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -145,51 +144,43 @@ private PrePopulatedValidationSupport loadIgs(final FhirContext ctx) {
 
         IParser parser = ctx.newJsonParser();
```
Bingjiling, line 145: Why is `ctx` removed as method input param if it's used here?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Validator.java
```diff
@@ -145,51 +144,43 @@ private PrePopulatedValidationSupport loadIgs(final FhirContext ctx) {
...
-                        String igResourcePath = indexFile.getPath().replace(".index.json", file.filename);
+                        String igResourcePath = indexFile.getKey().replace(".index.json", file.filename);
                         log.info("loading {}", igResourcePath);
-                        ResourceList resourcesWithPath = allFiles.getResourcesWithPath(igResourcePath);
+                        List<IGObject> resourcesWithPath = this.implementationGuidesResources.stream()
+                            .filter(resource -> resource.getKey().startsWith(igResourcePath))
+                            .collect(Collectors.toList());
                         if (resourcesWithPath.isEmpty()) {
-                            throw new RuntimeException("The following file is declared in .index.json but does not exist: " + igResourcePath);
+                            throw new RuntimeException("The following file is declared in .index.json but does not exit: " + igResourcePath);
```
Bingjiling, line 158: Typo

floofydoug:
Is there a reason why this PR was never merged? 

I have tried following the implementation guide readme for US Core  5.0.1, Core 4.0.0, and Core 3.1.1, of which make the hapiValidator deployment too large and cannot be deployed.

PR #501, opened 2021-10-27
chore: update serverless versions

rsmayda:
Issue #, if available:

Description of changes:
- Update serverless versions

Checklist:

- Ran all custom integration tests; did not run crucible tests
- Tested sls-offline and it still works

### DIFF
--- a/.github/workflows/cfn-analysis.yml
+++ b/.github/workflows/cfn-analysis.yml
@@ -26,11 +26,9 @@ jobs:
           node-version: 14
       - name: Install npm dependencies
         run: yarn install
-      - name: Install serverless
-        run: npm install -g serverless@2.64.1
       - name: Package main sls into cfn for analysis
         run: |
-          serverless package --stage dev --region us-west-2 --useHapiValidator true --conceal
+          yarn run serverless-package --stage dev --region us-west-2 --useHapiValidator true --conceal
       - name: cfn_nag on main serverless
         uses: stelligent/cfn_nag@master
         with:
@@ -40,7 +38,7 @@ jobs:
         run: |
           cd auditLogMover
           yarn install
-          serverless package --stage dev --region us-west-2 --conceal
+          yarn run serverless-package --stage dev --region us-west-2 --conceal
       - name: cfn_nag on auditLog serverless
         uses: stelligent/cfn_nag@master
         with:
--- a/.github/workflows/cfn-analysis.yml
+++ b/.github/workflows/cfn-analysis.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install npm dependencies
         run: yarn install
       - name: Install serverless
-        run: npm install -g serverless
+        run: npm install -g serverless@2.64.1
       - name: Package main sls into cfn for analysis
         run: |
           serverless package --stage dev --region us-west-2 --useHapiValidator true --conceal
@@ -35,7 +35,7 @@ jobs:
         uses: stelligent/cfn_nag@master
         with:
           input_path: .serverless/cloudformation-template-update-stack.json
-          extra_args: -o json  
+          extra_args: -o json
       - name: Package auditLogMover sls into cfn for analysis
         run: |
           cd auditLogMover
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -87,7 +87,7 @@ jobs:
       - name: Compile IGs
         run: yarn run compile-igs
       - name: Install serverless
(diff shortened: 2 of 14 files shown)


.github/workflows/cfn-analysis.yml
```diff
@@ -27,15 +27,15 @@ jobs:
       - name: Install npm dependencies
         run: yarn install
       - name: Install serverless
-        run: npm install -g serverless
+        run: npm install -g serverless@2.64.1
```
carvantes: we should move away from installing serverless globally. It is not needed and can cause confusion if the version of the global serverless is different from the local one.

the next step can run as 
```
./node_modules/bin/serverless package ...
```
or if we add a script on package.json it could be:

```
yarn run serverless package ...
```

We don't need to change it as part of this PR though.

carvantes (approved):

.github/workflows/cfn-analysis.yml
rsmayda: I like this change I will throw it in here as well

.github/workflows/deploy.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
       - name: Compile IGs
         run: yarn run compile-igs
       - name: Install serverless
-        run: npm install -g serverless
+        run: npm install -g serverless@2.64.1
```
rsmayda, line 90: I wasn't sure how to not do this for the java serverless file as there isn't scripts

scripts/s3-dist.sh
```diff
@@ -31,7 +31,7 @@ function cleanup(){
 }
 
 echo Installing dependencies
-npm install -g serverless && npm install
+npm install -g serverless@2.64.1 && npm install
```
carvantes, line 34: I think that line 43 below will not work without serverless globally installed.

docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM node:14.4
 
 # Required dependencies
-RUN apt-get update && apt-get install -yq python3-pip less && pip3 install boto3 && curl -fsSL --compressed  "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o /tmp/awscliv2.zip && unzip -q /tmp/awscliv2.zip -d /tmp && /tmp/aws/install && npm install serverless -g
+RUN apt-get update && apt-get install -yq python3-pip less && pip3 install boto3 && curl -fsSL --compressed  "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o /tmp/awscliv2.zip && unzip -q /tmp/awscliv2.zip -d /tmp && /tmp/aws/install && npm install serverless@2.64.1 -g
```
carvantes, line 4: loos like keeping global serverless here is needed since it runs the `serverless.sh` file

Maybe we should update `serverless.sh` to also not call global serverless. It's a bit of a chore since we also need to update the docs, so maaybe it's ok to do it later. AFAIK the `serverless.sh` file is not referenced anywhere else, so it should be safe to not update it right now

carvantes (commented):
I think that we also need 

https://github.com/carvantes/fhir-works-on-aws-deployment/blob/a9aebcc182255d305327463b1b2e0f7a463bad95/scripts/install.sh#L300

docker/Dockerfile
rsmayda: sounds good I will leave it in for now

scripts/s3-dist.sh
rsmayda: Agreed I will leave it in but pinned to 2.64.1

scripts/install.sh
```diff
@@ -297,12 +297,12 @@ fi
 
 echo -e "\n\nFHIR Works is deploying. A fresh install will take ~20 mins\n\n"
 ## Deploy to stated region
-yarn run serverless deploy --region $region --stage $stage || { echo >&2 "Failed to deploy serverless application."; exit 1; }
+yarn run serverless-deploy --region $region --stage $stage || { echo >&2 "Failed to deploy serverless application."; exit 1; }
```
carvantes, line 300: my bad, I didn't saw the yarn run. This file didn't need any updates since `yarn run serverless` effectively runs `serverless` from `node_modules/.bin`. The update is still correct though


https://classic.yarnpkg.com/lang/en/docs/cli/run/

rsmayda: Yeah I thought the same but felt it was better to have it routed via our scripts

carvantes (approved):

Issue #498, opened 2021-10-20
[Misc] Got ResourceNotFoundException error at audit-log-mover-poc-exportCloudwatchLogs


rhinozD:
**What's on your mind?**
I've got this error when I check audit-log-mover-poc-exportCloudwatchLogs LogGroup. 
`ERROR	Failed to kick off all export tasks ResourceNotFoundException: The specified log group does not exist.`

Is this a bug or just I did mistake somewhere?
I found that the log group which is exported here not exist in CloudWatch.
```
      CloudwatchExecutionLogGroup:
        Description: Cloudwatch Execution log group for storing request/responses for auditing purposes
        Value: !Join ['', ['API-Gateway-Execution-Logs_', !Ref ApiGatewayRestApi, '/', !Ref Stage]]
        Export:
          Name: !Join ['-', [CloudwatchExecutionLogGroup, !Ref Stage]]
```
[](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L794-L798)
and the API Gateway executionLogging is disabling.
`executionLogging: false`
[](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/serverless.yaml#L78)
**Versions (please complete the following information):**
 - v3.1.0

maghirardelli:
Hello @rhinozD,

Thank you for bringing this to our attention. We believe this is a bug and are looking into this now. We'll work on the fix soon.

Thanks,
Marianna

carvantes:
Hi,

This issue was fixed on https://github.com/awslabs/fhir-works-on-aws-deployment/pull/503. 

Note that after applying the fix, new logs will be exported to S3 as usual, but old logs that were not exported will remain in CloudWatch. You can move the old logs to s3 with the `create-export-task` API: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/S3ExportTasks.html

PR #252, opened 2021-03-17
chore: update log settings

carvantes:
Description of changes:
- Increase log retention time
- Update access log format and disable execution logs similar to https://github.com/awslabs/service-workbench-on-aws/pull/267

Checklist:

* [x] Have you successfully deployed to an AWS account with your changes?
* [ ] Have you written new tests for your core changes, as applicable?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #503, opened 2021-10-28
fix: Fix CloudWatch LogGroup name for auditLogMover

jn1119:
Issue #, if available: [Issue 498](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/498)

Description of changes:

* As part of [PR-252](https://github.com/awslabs/fhir-works-on-aws-deployment/pull/252) we changed the API Gateway log group from `Execution` to `Access` which meant that log group was not managed by APIGateway. With that change our new log group was managed by serverless but we didn't change our auditLogMover to point to the new log group. With this change we are ensuring that auditLogMover uses the correct log group when moving logs to S3.

More information on Access vs Execution logs [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-logging.html)

Testing done:

* Verified that auditLogMover is able to launch the export job successfully
```
Response
{
  "daysExported": [
    "2021-10-21"
  ],
  "message": "Successfully kicked off export tasks"
}
``` 

Checklist:

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -791,6 +791,16 @@ resources:
         Condition: isDev
         Description: App client id for the provisioning ES Kibana users.
         Value: !Ref KibanaUserPoolClient
+      CloudwatchAccessLogGroup:
+        Description: Cloudwatch Access log group for storing request/responses for auditing purposes
+        Value: !Ref ApiGatewayLogGroup
+        Export:
+          Name: !Join ['-', [CloudwatchAccessLogGroup, !Ref Stage]]
+      CloudwatchAccessLogGroupArn:
+        Description: Cloudwatch Access log group ARN for storing request/responses for auditing purposes
+        Value: !GetAtt ApiGatewayLogGroup.Arn
+        Export:
+          Name: !Join ['-', [CloudwatchAccessLogGroup, !Ref Stage, Arn]]
       CloudwatchExecutionLogGroup:
         Description: Cloudwatch Execution log group for storing request/responses for auditing purposes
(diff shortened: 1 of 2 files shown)


Bingjiling:
Would the log mover be able to account for the previous logs that didn't get moved?

jn1119 (commented):
@Bingjiling  Note that this PR is only fixing the most immediate problem going forward. We still need to find a way to move existing cloudwatch logs data that wasn't moved to S3.

serverless.yaml
```diff
@@ -791,6 +791,16 @@ resources:
...
+        Description: Cloudwatch Access log group for storing request/responses for auditing purposes
+        Value: !Ref ApiGatewayLogGroup
+        Export:
+          Name: !Join ['-', [CloudwatchAccessLogGroup, !Ref Stage]]
+      CloudwatchAccessLogGroupArn:
+        Description: Cloudwatch Access log group ARN for storing request/responses for auditing purposes
+        Value: !GetAtt ApiGatewayLogGroup.Arn
+        Export:
+          Name: !Join ['-', [CloudwatchAccessLogGroup, !Ref Stage, Arn]]
       CloudwatchExecutionLogGroup:
```
rsmayda, line 804: Are these still of value to output?

rsmayda (approved):

serverless.yaml
jn1119: Still need `CloudwatchExecutionLogGroup` since CloudFormation will still think auditLogger depends on it. Once this template and auditLogger's new template is executed we can remove this output which can be done in a later release.

PR #522, opened 2021-12-13
chore: add unit test coverage and enforcement

ssvegaraju:
Issue #, if available:
FHIR-545

Description of changes:
Added workflow for testing code cov and set limit to 0 initially to see what the coverage is at before choosing a limit. 

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,17 @@
+#
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+name: Code Coverage Report
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2 # Check out your repository
+    - run: yarn install # Install dependencies
+    - run: yarn test --coverage # Run test
+    - run: bash <(curl -s https://codecov.io/bash) # Upload to Codecov
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -5,7 +5,7 @@
 
 name: Code Coverage Report
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


nguyen102:
You can check the code coverage by running `yarn test-coverage` locally.

I had to run this to check for the code cov of FHIR repos a couple of weeks ago. You can see my results [here](https://quip-amazon.com/8cKYAMzb0tqM/Chamonix-How-We-Work#temp:C:MTZ725c071e2d4db47a9fcd7d987)

codecov.yml
```diff
@@ -0,0 +1,19 @@
+coverage:
+  round: down
+  precision: 2
+  range: '0...100'
+  status:
+    project:
+      default:
+        # basic
+        target: 0%
```
nguyen102: In the How We Work, I recommended 90% for this. All of our FHIR Works repo have more than 90% coverage.

ssvegaraju: That's super helpful, thanks! Was originally just trying to see what coverage we were at but it didn't occur to me to just use yarn.

codecov.yml
```diff
@@ -1,12 +1,12 @@
...
   round: down
   precision: 2
-  range: '0...100'
+  range: '90...100'
   status:
     project:
       default:
         # basic
-        target: 0%
+        target: 00%
```
nguyen102: Should this be 90? That's the target we're aiming for.

.github/workflows/code-coverage.yml
```diff
@@ -5,7 +5,7 @@
 
 name: Code Coverage Report
 
-on: [push, pull_request]
+on: [push]
```
nguyen102: I think we want the code coverage to run on pull requests as well, so the developer knows if their code coverage meets the guideline before they merge to develop

codecov.yml
ssvegaraju: Yes, that's a typo 😅

.github/workflows/code-coverage.yml
ssvegaraju: That makes sense, I'll put that back in.

nguyen102 (approved):

PR #539, opened 2022-01-05
chore: add unit test coverage and enforcement

ssvegaraju:
Issue #, if available:

Description of changes:
Fixing CodeCov coverage workflow for this repo.


codecov.yml
```diff
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         # basic
-        target: 90%
+        target: 50%
```
carvantes: why 50%?

carvantes (dismissed):

codecov.yml
ssvegaraju: I think this was a holdover from a previous iteration, just updated it!

codecov.yml
```diff
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         # basic
-        target: 90%
+        target: 95%
```
carvantes: 95% seems too high. Looking at the report we are currently below that, at 94.84%.
I'd suggest using 90%

carvantes (approved):

PR #557, opened 2022-02-09
feat: subscriptionReaper

ssvegaraju:
Issue #, if available:

Description of changes:
Added a scheduled lambda that only runs if the flag for Subscriptions is set. Uses persistence.getActiveSubscriptions to filter out active subscriptions and persistence.deleteResource to delete expired ones.

**Testing**
- deployed successfully to an AWS account.
- posted a Subscription resource using Postman
- ran the subscriptionReaper Lambda from the lambda interface on the AWS Console
- verified that expected subscription resources appeared from the getActiveSubscriptions call, and were filtered correctly
- ensured that correct subscription resources were deleted based on end date
- ----
- redeployed on a multi-tenant deployment
- ensured all operations on subscriptions ran as expected with the same method as above

### DIFF
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -16,7 +16,7 @@ custom:
   config: ${file(serverless_config.json)}
   useHapiValidator: ${opt:useHapiValidator, 'false'}
   enableMultiTenancy: ${opt:enableMultiTenancy, 'false'}
-  enableSubscriptions: ${opt:enableSubscriptions, 'true'}
+  enableSubscriptions: ${opt:enableSubscriptions, 'false'}
   logLevel: ${opt:logLevel, 'error'}
   enableESHardDelete: ${opt:enableESHardDelete, 'false'}
   patientCompartmentFileV3: 'patientCompartmentSearchParams.3.0.2.json'
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -192,14 +192,17 @@ functions:
       ELASTICSEARCH_DOMAIN_ENDPOINT: !Join ['', ['https://', !GetAtt ElasticSearchDomain.DomainEndpoint]]
       NUMBER_OF_SHARDS: !If [isDev, 1, 3] # 133 indices, one per resource types
     
-    subscriptionReaper:
-      timeout: 30
-      runtime: nodejs14.x
-      description: 'Scheduled Lambda to remove expired Subscriptions'
-      handler: subscriptions/index.handler
-      events:
-        - schedule: rate(5 minutes)
-        - enable: ${self:custom.enableSubscriptions}
+  subscriptionReaper:
+    timeout: 30
+    runtime: nodejs14.x
+    description: 'Scheduled Lambda to remove expired Subscriptions'
+    handler: src/subscriptions/index.handler
+    events:
+      - schedule: rate(5 minutes)
+      - enable: ${self:custom.enableSubscriptions}
+    environment:
+      RESOURCE_TABLE: ${self:custom.resourceTableName}
+      ENABLE_MULTI_TENANCY: !Ref EnableMultiTenancy
 
 stepFunctions:
   stateMachines:
--- a/serverless.yaml
+++ b/serverless.yaml
@@ -16,7 +16,7 @@ custom:
   config: ${file(serverless_config.json)}
   useHapiValidator: ${opt:useHapiValidator, 'false'}
   enableMultiTenancy: ${opt:enableMultiTenancy, 'false'}
-  enableSubscriptions: ${opt:enableSubscriptions, 'false'}
+  enableSubscriptions: ${opt:enableSubscriptions, 'true'}
   logLevel: ${opt:logLevel, 'error'}
   enableESHardDelete: ${opt:enableESHardDelete, 'false'}
   patientCompartmentFileV3: 'patientCompartmentSearchParams.3.0.2.json'
@@ -196,6 +196,7 @@ functions:
     timeout: 30
     runtime: nodejs14.x
     description: 'Scheduled Lambda to remove expired Subscriptions'
+    role: SubscriptionReaperRole
     handler: src/subscriptions/index.handler
     events:
       - schedule: rate(5 minutes)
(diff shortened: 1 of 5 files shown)


src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ */
+import { DynamoDbDataService, DynamoDb } from 'fhir-works-on-aws-persistence-ddb';
+
+const enableMultiTenancy = process.env.ENABLE_MULTI_TENANCY === 'true';
+const dbService = new DynamoDbDataService(DynamoDb, false, {
+    enableMultiTenancy,
```
carvantes, line 10: Now that I think of it, we actually need 2 `DynamoDbDataService`. The reaper needs a ddb without `enableMultiTenancy` since it will be deleting subscriptions for all tenants so it must be able to fetch all subscriptions with `getActiveSubscriptions`. 
But a ddb with `enableMultiTenancy` is needed to call `dbService.deleteResource` in order to deleet tenant-specific resources

src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,37 @@
...
+        // example format of subscriptions: https://www.hl7.org/fhir/subscription-example.json.html
+        return await Promise.all(
+            subscriptions
+                .filter((s: Record<string, any>) => s.end && currentTime >= new Date(s.end))
+                .map(async (subscription) => {
+                    // delete the subscription as it has reached its end time
+                    return dbService.deleteResource({
+                        resourceType: subscription.resourceType,
+                        id: subscription.id,
+                    });
```
carvantes: I think that we need to pass the `tenantId` from the Subscription here. Otherwise you'll get an error if multi-tenancy is enabled.

src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,37 @@
...
+                }),
+        );
+    } catch (e: any) {
+        throw new Error(`Subscription Reaper failed! Error: ${e}`);
+    }
```
carvantes: It's probably fine to just let it crash since we are not truly handling the error. Actually, I think this logs error as a string so it won't show the stack trace. I suggest removing the catch

src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,37 @@
...
+    console.log('subscriptionReaper event', event);
+    try {
+        const subscriptions = await dbService.getActiveSubscriptions({});
+        const currentTime = new Date();
+        // filter out subscriptions without a defined end time.
+        // check if subscription is past its end date (ISO format)
+        // example format of subscriptions: https://www.hl7.org/fhir/subscription-example.json.html
+        return await Promise.all(
+            subscriptions
+                .filter((s: Record<string, any>) => s.end && currentTime >= new Date(s.end))
```
carvantes: I think we should consider the case where the end date is not in ISO format. Right now it will always be ignored as it is an `Invalid Date` that cannot be compared with currentTime.

I think this behavior it's ok, but maybe just add a log like "Skipping subscription {id} since end date is not in a valid format: {end.date}"

carvantes (changes requested):

src/subscriptions/subscriptionReaper.ts
ssvegaraju: That makes sense, I added another dbService and handled multitenant deployments as you commented below.

src/subscriptions/subscriptionReaper.ts
ssvegaraju: That makes sense, I removed the try catch block!

src/subscriptions/subscriptionReaper.ts
ssvegaraju: I refactored the filter to log in the case that the condition fails due to no end date or an invalid end date

serverless.yaml
```diff
@@ -191,6 +191,17 @@ functions:
     environment:
       ELASTICSEARCH_DOMAIN_ENDPOINT: !Join ['', ['https://', !GetAtt ElasticSearchDomain.DomainEndpoint]]
       NUMBER_OF_SHARDS: !If [isDev, 1, 3] # 133 indices, one per resource types
+    
+  subscriptionReaper:
```
carvantes, line 195: does this lambda has enough permissions? I think It needs permissions to query the new DDB "activeSubscriptions" index

src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,43 @@
...
+    // check if subscription is past its end date (ISO format)
+    // example format of subscriptions: https://www.hl7.org/fhir/subscription-example.json.html
+    return Promise.all(
+        subscriptions
+            .filter((s: Record<string, any>) => {
+                if (s.end && currentTime >= new Date(s.end)) {
+                    return true;
+                }
+                console.log(`Skipping subscription ${s.id} since the end date is not in a valid format: ${s.end}`);
+                return false;
```
carvantes: I don't think this works as expected. 

Looks like a valid date in the future will return false and print the "date is not in a valid format"

carvantes (changes requested):

src/subscriptions/subscriptionReaper.test.ts
```diff
@@ -0,0 +1,90 @@
...
+                id: 'sub2',
+                status: 'requested',
+                end: '2121-01-01T00:00:00Z',
+            },
+        ];
+        const mockGetActiveSubscriptions = jest.fn();
+        const mockDeleteResource = jest.fn();
+        DynamoDbDataService.prototype.getActiveSubscriptions = mockGetActiveSubscriptions;
+        DynamoDbDataService.prototype.deleteResource = mockDeleteResource;
+        mockGetActiveSubscriptions.mockReturnValue(Promise.resolve(subResource));
```
Bingjiling: You can use `mockResolvedValue()` and `mockRejectedValue()` for async functions. Also it's a good idea to mock the value once only so it does not bring unintended side effect to other tests(e.g. `mockResolvedValueOnce()`, `mockReturnValueOnce()` )

serverless.yaml
ssvegaraju: From my testing it did not have enough permissions, so I added a role for it in serverless.yaml.

src/subscriptions/subscriptionReaper.ts
ssvegaraju: I revised the logic to be more clear about the different cases here.

src/subscriptions/subscriptionReaper.test.ts
ssvegaraju: That's a good point, I removed the Promise.resolve and used the mockResolveValueOnce method instead.

src/subscriptions/subscriptionReaper.ts
```diff
@@ -0,0 +1,46 @@
...
+    console.log('subscriptionReaper event', event);
+    const subscriptions = await dbService.getActiveSubscriptions({});
+    const currentTime = new Date();
+    // filter out subscriptions without a defined end time.
+    // check if subscription is past its end date (ISO format)
+    // example format of subscriptions: https://www.hl7.org/fhir/subscription-example.json.html
+    return Promise.all(
+        subscriptions
+            .filter((s: Record<string, any>) => {
+                // if s.end is undefined, new Date(s.end) will throw an error
```
carvantes: Maybe remove this comment. I don't think an error is **thrown** here

carvantes (approved):

PR #558, opened 2022-02-09
feat: Rest hook Lambda

Bingjiling:
Issue #, if available:

Description of changes:

* Create Lambda that sends rest-hook POST and PUT notifications 
* Tested with event `{"subscriptionId": "123456","channelType":"rest-hook","endpoint": "https://postman-echo.com/post","channelHeader": ["test:poo"],"matchedResource": {"id":"1234567","resourceType": "Patient","versionId": "string","lastUpdated": "string"}}` from SNS publish
* Note - The above event can not be parsed correctly if there's new line character(/n), will need to further verify the parsing behavior once we integrate with the subscription Matcher Lambda

Checklist:


cloudformation/subscriptions.yaml
```diff
@@ -96,4 +96,19 @@ Resources:
...
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service: 'lambda.amazonaws.com'
+            Action: 'sts:AssumeRole'
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
```
Bingjiling: WIP - Shrinking permission scope

package.json
```diff
@@ -62,10 +63,9 @@
     "jest-circus": "^26.6.3",
     "jest-mock-extended": "^1.0.8",
     "jsonwebtoken": "^8.5.1",
-    "lodash": "^4.17.21",
     "prettier": "^2.4.1",
     "qs": "^6.10.1",
-    "serverless": "2.64.1",
+    "serverless": "2.72.2",
```
carvantes: Is this needed? 
If not, It's probably better not to update this on a feature branch

serverless.yaml
```diff
@@ -192,6 +192,17 @@ functions:
       ELASTICSEARCH_DOMAIN_ENDPOINT: !Join ['', ['https://', !GetAtt ElasticSearchDomain.DomainEndpoint]]
       NUMBER_OF_SHARDS: !If [isDev, 1, 3] # 133 indices, one per resource types
 
+  subscriptionsRestHook:
+    timeout: 700
```
carvantes: I suggest a much smaller timeout. Maybe 10 seconds

subscriptions/restHook.ts
```diff
@@ -0,0 +1,127 @@
...
+import { makeLogger } from 'fhir-works-on-aws-interface';
+import { SubscriptionMatchMessage } from './types';
+import getAllowListedSubscriptionEndpoints from './allowList';
+import ensureAsyncInit from '../src/index';
+
+const logger = makeLogger();
+
+const SINGLE_TENANT_ALLOW_LIST_KEY = 'SINGLE_TENANT_ALLOW_LIST_KEY';
+
+export interface AllowListInfo {
```
carvantes: can you move all the functions related to the allow list to a separate file? I feel that this file is bit too big

subscriptions/restHook.ts
```diff
@@ -0,0 +1,127 @@
...
+                return headerMap[allowedEndpoint.toString()];
+            }
+            if (allowedEndpoint === endpoint) {
+                return headerMap[allowedEndpoint];
+            }
+        }
+        throw new Error(`Endpoint ${endpoint} is not allow listed.`);
+    };
+
+    if (process.env.ENABLE_MULTI_TENANCY === 'true') {
```
carvantes: We should avoid accessing env variables deep within our code (I know FWoA does that a lot but it's a bad practice.). It makes it hard to know how the code behaves(e.g. looking at the index file it is hard to tell that `
exports.handler = sendRestHookNotification;` behaves differently with multi-tenancy)

I sugest to pass enableMultitenancy as a param instead. You may need to refactor into a class to "remember" the value of the env variable
```ts
const restHookHandler = new RestHookHandler({ enableMultitenancy: process.env.ENABLE_MULTI_TENANCY === 'true' });

export async function handleSQSevent(event: any) {
    return restHookHandler.sendRestHookNotification(event);
}
```

although passing the param on all function calls works as well
```ts
export async function handleSQSevent(event: any) {
    sendRestHookNotification(event, { enableMultitenancy: process.env.ENABLE_MULTI_TENANCY === 'true' });
}
```

subscriptions/restHook.ts
```diff
@@ -0,0 +1,127 @@
...
+    };
+
+    allowListHeader.forEach((header) => mergeHeader(header));
+    channelHeader.forEach((header) => mergeHeader(header));
+    return mergedHeader;
+};
+
+const allowListPromise: Promise<{ [key: string]: AllowListInfo }> = getAllowListInfo();
+
+export const sendRestHookNotification = async (event: any): Promise<any> => {
```
carvantes: you can use the SQSEvent type from `aws-lambda`.
`aws-lambda` has types for all sorts of AWS events. It's convenient to use it instead of trying to figure out the event schema from documentation

e.g. https://github.com/awslabs/fhir-works-on-aws-search-es/blob/280ea2bc5b76df948586b55153221aeb82bdb1d0/src/StreamSubscriptionMatcher/StreamSubscriptionMatcher.ts#L7

serverless.yaml
carvantes: looks like `700` fails with: 
`Resource handler returned message: "Invalid request provided: Queue visibility timeout: 30 seconds is less than Function timeout: 700 seconds `

package.json
Bingjiling: I updated it for a bug I ran into deploying CloudFormation with SNS, I'll double check if it's needed.

serverless.yaml
Bingjiling: Emmmm I didn't ran into this error when deploying, will update timeout.

subscriptions/restHook.ts
Bingjiling: Sure.

subscriptions/restHook.ts
Bingjiling: Good point. I'll move to refactor into a class.

subscriptions/restHook.ts
Bingjiling: Nice! I'll check if it works. I did ran into some issues parsing the event, and from the docs I read, it mostly came from SNS stringify the message. Hopefully this will help.

src/subscriptions/allowListUtil.ts
```diff
@@ -0,0 +1,82 @@
...
+            if (allowedEndpoint === endpoint) {
+                return headerMap[allowedEndpoint];
+            }
+        }
+        throw new Error(`Endpoint ${endpoint} is not allow listed.`);
+    };
+
+    if (enableMultitenancy) {
+        if (tenantId) {
+            return getHeaders(allowListInfo[tenantId]);
```
carvantes: what happens if there are no entries in the allow list for the given `tenantId`?

I'm doing some tests and I get an error:
```
"errorType": "TypeError",
    "errorMessage": "Cannot destructure property 'allowList' of 'undefined' as it is undefined.",
    "stack": [
        "TypeError: Cannot destructure property 'allowList' of 'undefined' as it is undefined.",
        "    at getHeaders (/var/task/subscriptions/webpack:/Users/nestorba/Amazon/AWS/github/workspace/fhir-works-on-aws-deployment/subscriptions/restHook.ts:58:27)",
        "    at getAllowListHeaders (/var/task/subscriptions/webpack:/Users/nestorba/Amazon/AWS/github/workspace/fhir-works-on-aws-deployment/subscriptions/restHook.ts:73:20)",
        "    at /var/task/subscriptions/webpack:/Users/nestorba/Amazon/AWS/github/workspace/fhir-works-on-aws-deployment/subscriptions/restHook.ts:114:34",
        "    at Array.map (<anonymous>)",
        "    at Runtime.sendRestHookNotification [as handler] (/var/task/subscriptions/webpack:/Users/nestorba/Amazon/AWS/github/workspace/fhir-works-on-aws-deployment/subscriptions/restHook.ts:109:48)"
    ]
```

Maybe my dev setup is messed up, but please double check if this is a bug

Bingjiling: Thanks! I'll check on it.

carvantes (approved):

PR #570, opened 2022-02-17
feat: emit end to end latency metric from rest-hook Lambda

Bingjiling:
Issue #, if available:

Description of changes:

* Emit end to end latency metric from rest-hook Lambda

Screen Shot 2022-02-18 at 2 23 12 PM

Checklist:

### DIFF
--- a/src/subscriptions/restHookLambda/restHook.ts
+++ b/src/subscriptions/restHookLambda/restHook.ts
@@ -2,6 +2,8 @@ import axios from 'axios';
 import { makeLogger } from 'fhir-works-on-aws-interface';
 import { SQSEvent } from 'aws-lambda';
 import { SubscriptionNotification } from 'fhir-works-on-aws-search-es';
+import { metricScope, Unit } from 'aws-embedded-metrics';
+import moment from 'moment';
 import ensureAsyncInit from '../../index';
 import { AllowListInfo, getAllowListHeaders } from './allowListUtil';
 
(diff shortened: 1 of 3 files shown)


src/subscriptions/restHookLambda/restHook.ts
```diff
@@ -2,6 +2,8 @@ import axios from 'axios';
 import { makeLogger } from 'fhir-works-on-aws-interface';
 import { SQSEvent } from 'aws-lambda';
 import { SubscriptionNotification } from 'fhir-works-on-aws-search-es';
+import { metricScope, Unit } from 'aws-embedded-metrics';
+import moment from 'moment';
```
carvantes: I'd suggest not using `moment`

The difference in milliseconds is easy to calculate using the available Date functions
```
Date.now() - (new Date('2022-02-17T20:33:57.021Z')).getTime()
```

Also, moment is sort of deprecated nowadays: https://github.com/moment/moment/#project-status

src/subscriptions/restHookLambda/restHook.ts
```diff
@@ -19,15 +21,29 @@ const mergeRequestHeaders = (allowListHeader: string[], channelHeader: string[])
...
     allowListHeader.forEach((header) => mergeHeader(header));
     channelHeader.forEach((header) => mergeHeader(header));
     return mergedHeader;
 };
 
+/**
+ * Push latency metric to CloudWatch
+ * @param messages
+ */
+const pushLatencyMetric = metricScope((metrics) => async (messages: SubscriptionNotification[]): Promise<void> => {
```
carvantes: nit: rename to something like `recordLatencyMetric` or ``logLatencyMetric``. I think that "push" sounds misleading

carvantes (changes requested):

src/subscriptions/restHookLambda/restHook.ts
Bingjiling: Wow, I didn't know moment is getting deprecated. Will update to Date.

src/subscriptions/restHookLambda/restHook.ts
```diff
@@ -19,15 +20,29 @@ const mergeRequestHeaders = (allowListHeader: string[], channelHeader: string[])
...
+ * Push latency metric to CloudWatch
+ * @param messages
+ */
+const logLatencyMetric = metricScope((metrics) => async (messages: SubscriptionNotification[]): Promise<void> => {
+    const currentTime = new Date().getTime();
+    messages.forEach((message: SubscriptionNotification) => {
+        metrics.putMetric(
+            'SubscriptionEndToEndLatency',
+            (currentTime - new Date(message.matchedResource.lastUpdated).getTime()) / 1000,
+            Unit.Seconds,
```
carvantes: It's almost always better to emit metrics in the smallest unit available. In this case Milliseconds.

When storing milliseconds you can always divide by 1000 on the graphs if you want to see seconds. But storing the metric as seconds destroys the milliseconds information.

carvantes (changes requested):

carvantes (approved):

PR #575, opened 2022-02-22
test: add Subscription reaper tests

ssvegaraju:
Issue #, if available:

Description of changes:
**This PR will be continuously updated with each set of integ tests!**
Added tests for the two cases of creating a subscription with an expiration date in the near future and making sure it is deleted by the subscription reaper. Also added a fix for a bug related to the subscription reaper and tenancy.
- This was tested by running `yarn run int-test` and ensuring that the tests passed as expected. I also manually tested the subscription reaper lambda by creating resources with different tenancy permutations and making sure that the current deployment works for both situations.

Checklist:


integration-tests/subscriptions.test.ts
```diff
@@ -39,4 +46,39 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
             console.log(x);
         });
     });
+
+    let client: AxiosInstance;
+    describe('test subscription creation and deletion', () => {
+        beforeAll(async () => {
+            client = await getFhirClient({ tenant: MULTI_TENANCY_ENABLED ? 'tenant1' : undefined });
```
carvantes: `getFhirClient` already has the logic to default to tenant1 when multi-tennacy is enabled. There's no need to repeat that here.

In general most tests simply call `getFhirClient()` unless they need a specific role or a specific tenant( for tests that involve many tenants or for tenant isolation tests)

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/401e97d9788a8998fbea707b80d5a0dd1aede87c/integration-tests/utils.ts#L79

https://github.com/awslabs/fhir-works-on-aws-deployment/blob/401e97d9788a8998fbea707b80d5a0dd1aede87c/integration-tests/utils.ts#L110

integration-tests/subscriptions.test.ts
```diff
@@ -39,4 +46,39 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+        });
+
+        test('creation of almost expiring subscription should be deleted by reaper', async () => {
+            // OPERATE
+            const subscriptionResource = {
+                resourceType: 'Subscription',
+                status: 'requested',
+                // get a time 10 seconds (10000 ms) in the future
+                end: new Date(new Date().getTime() + 10000).toISOString(),
+                reason: 'Monitor Patients for Organization 123',
```
carvantes: nit: change the reason to match the `criteria` field

integration-tests/subscriptions.test.ts
```diff
@@ -39,4 +46,39 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+                    endpoint: 'https://customer-endpoint.com',
+                    payload: 'application/fhir+json',
+                    header: ['Authorization: Bearer secret-token-abc-123'],
+                },
+            };
+            const postSubResult = await client.post('Subscription', subscriptionResource);
+            expect(postSubResult.status).toEqual(201); // ensure that the sub resource is created
+            const subResourceId = postSubResult.data.id;
+            // wait 4 min until reaper has had a chance to run
+            // (waiting 5 min would hit the timeout value above)
```
carvantes: just increase the timeout if needed.

```
jest.setTimeout(700_000);
```

integration-tests/subscriptions.test.ts
```diff
@@ -39,4 +46,39 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            // it may be better to do a loop checking every minute or so or increasing the timeout
+            await waitForMs(4 * 60 * 1000);
+            await expect(client.get(`Subscription/${subResourceId}`)).rejects.toMatchObject({
+                response: { status: 404 },
+            });
```
carvantes: let's check every 30 seconds instead of waiting for a flat 4 minutes (also, 4 minutes is not enough to guarantee that the reaper runs)

you can use `waitForExpect`, a library we already use that does the checks in a loop. 

Here's an example: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/c9a2c78c560a67e69277df572e8d1c9f1183aa88/integration-tests/implementationGuides.test.ts#L249-L263

something like: 
```ts
await waitForExpect(
            () => {
                console.log('Checking if Subscription/111 has already been deleted');
                expect(client.get(`Subscription/111`)).rejects.toMatchObject({
                    response: { status: 404 },
                });
            },
            360_000,
            30_000,
        );
```

src/subscriptions/reaperLambda/subscriptionReaper.ts
```diff
@@ -26,10 +30,12 @@ const reaperHandler = async (dbService: DynamoDbDataService, dbServiceWithTenanc
             })
             .map(async (subscription) => {
                 // delete the subscription as it has reached its end time
+                // multi-tenant deployments have the .id as {tenantid}|{id}
+                // eslint-disable-next-line no-underscore-dangle
+                const id = multiTenancyEnabled ? subscription._id : subscription.id;
```
carvantes, line 35: This fix looks good. Ideally you could have done it in a separate PR and have it merged right away. The integ tests take much longer to write/review

carvantes (changes requested):

integration-tests/subscriptions.test.ts
carvantes: actually, it was missing async await before
```ts
await waitForExpect(
            async () => {
                console.log('Checking if Subscription/111 has already been deleted');
                await expect(client.get(`Subscription/111`)).rejects.toMatchObject({
                    response: { status: 404 },
                });
            },
            360_000,
            30_000,
        );
```

carvantes (approved):

PR #577, opened 2022-02-24
test: add tests for tenant isolation of subscriptions

ssvegaraju:
Issue #, if available:
FHIR -764

Description of changes:
Added tests for tenant isolation on subscriptions. The only modifications made to the other tests were to extract common elements such as subscription resource construction.

Tested by running `yarn int-test` and ensuring through the AWS Console that the lambda function ran as expected and the notifications were present in the DDB Table.

Checklist:

### DIFF
--- a/integration-tests/subscriptions.test.ts
+++ b/integration-tests/subscriptions.test.ts
@@ -6,15 +6,21 @@
 
 import * as AWS from 'aws-sdk';
 import { AxiosInstance } from 'axios';
-import { clone } from 'lodash';
+import { v4 } from 'uuid';
 import waitForExpect from 'wait-for-expect';
 import { SubscriptionsHelper } from './SubscriptionsHelper';
-import { getFhirClient } from './utils';
+import { getFhirClient, randomSubscription } from './utils';
 
 jest.setTimeout(700_000);
 
-const { SUBSCRIPTIONS_ENABLED, SUBSCRIPTIONS_NOTIFICATIONS_TABLE, SUBSCRIPTIONS_ENDPOINT, API_AWS_REGION } =
-    process.env;
+const {
+    SUBSCRIPTIONS_ENABLED,
+    SUBSCRIPTIONS_NOTIFICATIONS_TABLE,
+    SUBSCRIPTIONS_ENDPOINT,
+    API_AWS_REGION,
+    SUBSCRIPTIONS_API_KEY,
+    MULTI_TENANCY_ENABLED,
+} = process.env;
 
 if (API_AWS_REGION === undefined) {
     throw new Error('API_AWS_REGION environment variable is not defined');
@@ -27,25 +33,13 @@ test('empty test placeholder', () => {
 });
 
 if (SUBSCRIPTIONS_ENABLED === 'true') {
-    if (!SUBSCRIPTIONS_ENDPOINT) {
+    if (SUBSCRIPTIONS_ENDPOINT === undefined) {
         throw new Error('SUBSCRIPTIONS_ENDPOINT environment variable is not defined');
     }
+    if (SUBSCRIPTIONS_API_KEY === undefined) {
+        throw new Error('SUBSCRIPTIONS_API_KEY environment variable is not defined');
+    }
     let client: AxiosInstance;
-
-    const subscriptionResource = {
-        resourceType: 'Subscription',
-        status: 'requested',
-        // get a time 10 seconds (10000 ms) in the future
-        end: new Date(new Date().getTime() + 10000).toISOString(),
-        reason: 'Monitor Patients with name Smith',
-        criteria: 'Patient?name=Smith',
-        channel: {
-            type: 'rest-hook',
-            endpoint: SUBSCRIPTIONS_ENDPOINT!,
-            payload: 'application/fhir+json',
-            header: [`x-api-key: ${process.env.SUBSCRIPTIONS_API_KEY}`],
-        },
-    };
 
     describe('FHIR Subscriptions', () => {
         let subscriptionsHelper: SubscriptionsHelper;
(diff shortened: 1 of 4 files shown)


integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
+    const subscriptionResource = {
+        resourceType: 'Subscription',
+        status: 'requested',
+        // get a time 10 seconds (10000 ms) in the future
+        end: new Date(new Date().getTime() + 10000).toISOString(),
+        reason: 'Monitor Patients with name Smith',
+        criteria: 'Patient?name=Smith',
+        channel: {
+            type: 'rest-hook',
+            endpoint: SUBSCRIPTIONS_ENDPOINT!,
```
carvantes: explicitly fail the test suit if `SUBSCRIPTIONS_ENDPOINT` is undefined (just like `API_AWS_REGION` in L19). It makes it a lot easier to find out why your local tests are not working

integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
+            payload: 'application/fhir+json',
+            header: [`x-api-key: ${process.env.SUBSCRIPTIONS_API_KEY}`],
```
carvantes: explicitly fail the test suit if `SUBSCRIPTIONS_API_KEY` is undefined

integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
             console.log(x);
         });
+
+        if (process.env.MULTI_TENANCY_ENABLED === 'true') {
+            test('tenant isolation', async () => {
+                const clientAnotherTenant = await getFhirClient({ tenant: 'tenant2' });
+                // tenant 1 creates a subscription
+                const subResource = clone(subscriptionResource);
+                // make sure the end date isn't caught by the reaper before the test completes
+                subResource.end = new Date(new Date().getTime() + 100000).toISOString();
```
carvantes: I suggest increasing time to at least 10 minutes (`600_000`). There's no need for it to be such a tight time window

integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
         test('test', async () => {
             const x = await subscriptionsHelper.getNotifications('/lala');
             console.log(x);
         });
+
+        if (process.env.MULTI_TENANCY_ENABLED === 'true') {
+            test('tenant isolation', async () => {
+                const clientAnotherTenant = await getFhirClient({ tenant: 'tenant2' });
+                // tenant 1 creates a subscription
+                const subResource = clone(subscriptionResource);
```
carvantes: I suggest to create a function that generates Subscriptions instead of cloning the same one over and over. An example function to generate test Patients is: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/401e97d9788a8998fbea707b80d5a0dd1aede87c/integration-tests/utils.ts#L137

I know some tests do the clone stuff, but it's not a good practice and we try not to d that anymore.

integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
+                            family: 'Smith',
+                        },
+                    ],
+                };
+                // post matching resource on another tenant
+                const postPatientResult = await clientAnotherTenant.post('Patient', resourceThatMatchesSubscription);
+                expect(postPatientResult.status).toEqual(201);
+                // give SLA of 20 seconds for notification to be placed in ddb table
+                await new Promise((r) => setTimeout(r, 20000));
+                // make sure no notification was receieved for first tenant
```
carvantes: We need to wait more time. Subscriptions may take up to a minute to actually start sending notifications(due to the cache in the SubscriptionMatcher), also, we need leniency on the SLA of 20 seconds (a system with steady traffic sends notifications well below 20 seconds, but with very little traffic (like integ tests envs) subscriptions face cold starts + batching windows in stream/sqs processing that make them slower)

I'd say update it to 2 minutes waiting.

integration-tests/subscriptions.test.ts
```diff
@@ -31,44 +27,81 @@ test('empty test placeholder', () => {
...
+                const notifications = await subscriptionsHelper.getNotifications(
+                    `/Patient/${postPatientResult.data.id}`,
```
carvantes: This can get mixed up with Subscriptions created by  other tests. 

I suggest you add a UUID to the endpoint in all test Subscriptions, so that you can lookup in DDB for the path with the exact same UUID.

```ts
endpoint: `${SUBSCRIPTIONS_ENDPOINT}/<uuid>`
```

carvantes (changes requested):

integration-tests/subscriptions.test.ts
ssvegaraju: This happens on L30-L32, but I can move it next to the API_AWS_REGION or inside the test itself to make it more clear.

carvantes: my bad. Still, it's probably fine to move it to the top of the file since other tests will use it. 
Also the `!` is not needed then on `SUBSCRIPTIONS_ENDPOINT!` if there is already an undefined check

integration-tests/subscriptions.test.ts
ssvegaraju: Made this the default time window for subscriptions created in utils.ts, and just changed the end time for the one test that needed things to expire quickly.

integration-tests/subscriptions.test.ts
ssvegaraju: Gotcha, I will avoid clone in the future. For now I've created a rather basic method to create subscriptions that takes in a uuid in utils.ts

integration-tests/subscriptions.test.ts
ssvegaraju: Updated to 2 min!

integration-tests/subscriptions.test.ts
ssvegaraju: Used the uuid package to generate uuids for each test that generates subscriptions so no test interferes with another.

carvantes (approved):

PR #578, opened 2022-02-24
test: add end to end tests for subscriptions

ssvegaraju:
Issue #, if available:

Description of changes:
Added end to end tests for empty notification and id notification scenarios. Tested using `yarn int-test` and making sure that the notification table and output looked as expected through the AWS Console.

Checklist:

### DIFF
--- a/integration-tests/subscriptions.test.ts
+++ b/integration-tests/subscriptions.test.ts
@@ -9,7 +9,7 @@ import { AxiosInstance } from 'axios';
 import { v4 } from 'uuid';
 import waitForExpect from 'wait-for-expect';
 import { SubscriptionsHelper } from './SubscriptionsHelper';
-import { getFhirClient, randomSubscription } from './utils';
+import { getFhirClient, randomSubscription, sleep } from './utils';
 
 jest.setTimeout(700_000);
 
@@ -74,11 +74,12 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
                 const subResource = randomSubscription(uuid);
                 const postResult = await client.post('Subscription', subResource);
                 expect(postResult.status).toEqual(201);
-                // post matching resource on another tenant
+                // post matching resource on another tenant, waiting 1 minute for subscription to enable
+                await sleep(60_000);
                 const postPatientResult = await clientAnotherTenant.post('Patient', resourceThatMatchesSubscription);
                 expect(postPatientResult.status).toEqual(201);
-                // give 2 minutes for notification to be placed in ddb table
-                await new Promise((r) => setTimeout(r, 120_000));
+                // give 1 minute for notification to be placed in ddb table
+                await sleep(60_000);
                 // make sure no notification was receieved for first tenant
                 const notifications = await subscriptionsHelper.getNotifications(
                     `${uuid}/Patient/${postPatientResult.data.id}`,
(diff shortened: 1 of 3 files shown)


integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
                 expect(notifications).toEqual([]);
             });
         }
+
+        test('end to end test with id notifications', async () => {
+            const uuid = v4();
+            const subResource = randomSubscription(uuid);
+            // 1. Create a Subscription.
```
carvantes, line 125: Can you add line breaks before each numbered step? It would make the test much more readable

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            const postSubscriptionResult = await client.post('Subscription', subResource);
+            expect(postSubscriptionResult.status).toBe(201);
+            // 2. Create/Update a resource that matches the subscription.
+            const postPatientResult = await client.post('Patient', resourceThatMatchesSubscription);
+            expect(postPatientResult.status).toEqual(201);
+            // 3. Verify that notifications are received
+            // give 2 minutes for notification to be placed in ddb table
```
carvantes: actually we need to wait one minute after creating the subscription but BEFORE posting the Patient that matches the subscription.(otherwise the Subscription may not be enabled yet due to caching)

I think that this needs to be fixed on tenant isolation tests as well

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            await new Promise((r) => setTimeout(r, 120_000));
```
carvantes: This is used several times, It'd be better to declare this helper function once. 
You can move an existing `sleep` fn to test utils and then use it everywhere: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/f122394c3f06ff380b51cbb513dc92f83a49347c/integration-tests/bulkExport.test.ts#L12

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            // make sure notification was receieved
+            let notifications = await subscriptionsHelper.getNotifications(`/${uuid}`);
+            expect(notifications).not.toEqual([]);
+            expect(notifications[0].httpMethod).toEqual('POST');
+            expect(notifications[0].body).toBeNull();
+            // get resources updated within three minutes ago
+            // now we search for our patient to make sure it was updated correctly
+            const searchResult = await client.get(
+                `${subResource.criteria}&_lastUpdated=gt${new Date(new Date().getTime() - 180_000).toISOString()}`,
+            );
```
carvantes: Not really needed. There are other existing tests for update/search

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            // 3. Verify that notifications are received
+            // give 2 minutes for notification to be placed in ddb table
+            await new Promise((r) => setTimeout(r, 120_000));
+            // make sure notification was receieved
+            let notifications = await subscriptionsHelper.getNotifications(
+                `/${uuid}/Patient/${postPatientResult.data.id}`,
+            );
+            expect(notifications).not.toEqual([]);
+            expect(notifications[0].httpMethod).toEqual('PUT');
+            expect(notifications[0].body).toBeNull();
```
carvantes: This would be faster using `waitForExpect`. That way you don't need to wait an entire minute.

Also, can you assert that the headers match?

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            // 4. Delete the Subscription
+            const deleteSubscriptionResult = await client.delete(`Subscription/${postSubscriptionResult.data.id}`);
+            expect(deleteSubscriptionResult.status).toEqual(200);
+            await new Promise((r) => setTimeout(r, 120_000));
```
carvantes: This wait can be only 1 minute

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
         }
+
+        test('end to end test with id notifications', async () => {
+            const uuid = v4();
+            const subResource = randomSubscription(uuid);
+            // 1. Create a Subscription.
+            const postSubscriptionResult = await client.post('Subscription', subResource);
+            expect(postSubscriptionResult.status).toBe(201);
+            // 2. Create/Update a resource that matches the subscription.
+            const postPatientResult = await client.post('Patient', resourceThatMatchesSubscription);
```
carvantes: We need to test Update as well.

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            });
+            expect(updatePatientResult.status).toEqual(200);
+            // test create:
+            const postAnotherPatientResult = await client.post('Patient', resourceThatMatchesSubscription);
+            expect(postAnotherPatientResult.status).toEqual(201);
+            // 6. Verify that notifications are no longer being sent
+            await new Promise((r) => setTimeout(r, 120_000));
+            notifications = await subscriptionsHelper.getNotifications(`/${uuid}/Patient/${postPatientResult.data.id}`);
+            // we still have the one notification from earlier in the test, but no more
+            expect(notifications.length).toEqual(1);
```
carvantes: I think that we also need to lookup notifications for `postAnotherPatientResult.data.id` and make sure that there are none.

integration-tests/subscriptions.test.ts
```diff
@@ -86,6 +86,90 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+        });
+
+        test('end to end test with empty notifications', async () => {
```
carvantes, line 188: There's a lot of repeated code for this test. I suggest ending this test at `// 3. Verify that notifications are received`.

There isn't any specific deletion behavior for empty notifications. It's ok for the test to only focus on asserting the contents of an "empty notification".

carvantes (changes requested):
Looks good. There are several minor changes to be made. 

Also, can you add tests for Subscription validation? those ones are easy, just post invalid Subscriptions (non-allowlisted endpoint, unsupported channel, invalid criteria) and assert that you get a 400 error

integration-tests/subscriptions.test.ts
```diff
@@ -65,27 +74,146 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+                response: { status: 400 },
+            });
+
+            // test for unsupported channel types (email, sms, and websocket)
+            subResource = randomSubscription(v4());
+            subResource.channel.type = 'email';
+            await expect(client.post('Subscription', subResource)).rejects.toMatchObject({
+                response: { status: 400 },
+            });
+            subResource.channel.type = 'SMS';
```
carvantes: this needs to be `sms` (case sensitive). Make sure that you are using the codes as they appear on https://fhir-ru.github.io/valueset-subscription-channel-type.html#expansion across this tests

Here you are getting a 404 because 'SMS' is an invalid code, so the validator (either HAPI or JSON schema) reject it, but what we want to test is the custom logic we wrote to reject valid channels that we don't support

integration-tests/subscriptions.test.ts
```diff
@@ -65,27 +74,146 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            });
+            expect(updatePatientResult.status).toEqual(200);
+
+            // 3. Verify that notifications are received
+            await waitForExpect(
+                async () => {
+                    const notifications = await subscriptionsHelper.getNotifications(
+                        `/${uuid}/Patient/${postPatientResult.data.id}`,
+                    );
+                    expect(notifications).not.toEqual([]);
```
carvantes: you can remove this line. The `length` check on the next line is sufficient

integration-tests/subscriptions.test.ts
```diff
@@ -65,27 +74,146 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+                    expect(notifications.length).toEqual(2); // one for create, one for update
+                    expect(notifications[0].httpMethod).toEqual('PUT');
+                    expect(notifications[0].body).toBeNull();
+                    expect(notifications[0].headers).toHaveProperty('x-api-key', SUBSCRIPTIONS_API_KEY);
```
carvantes, line 149: for completeness, can you also make this assertions for `notifications[1]`?

integration-tests/subscriptions.test.ts
```diff
@@ -65,27 +74,146 @@ if (SUBSCRIPTIONS_ENABLED === 'true') {
...
+            expect(updatePatientResult.status).toEqual(200);
+
+            // 3. Verify that notifications are received
+            await waitForExpect(
+                async () => {
+                    const notifications = await subscriptionsHelper.getNotifications(`/${uuid}`);
+                    expect(notifications).not.toEqual([]);
+                    expect(notifications[0].httpMethod).toEqual('POST');
+                    expect(notifications[0].body).toBeNull();
+                    expect(notifications[0].headers).toHaveProperty('x-api-key', SUBSCRIPTIONS_API_KEY);
```
carvantes, line 213: Same as above, let's assert the same for `notifications[1]`

integration-tests/subscriptions.test.ts
ssvegaraju: Gotcha, I've updated the test to the correct spelling and also updated websocket to match the standard as well.

carvantes (approved):
🚀

Issue #624, opened 2022-05-13
[Feature Request] Log when express server is started and stopped


Zambonilli:
**Is your feature request related to a problem? Please describe.**
Trying to debug when new instances of our fhirServer lambda are started and stopped is currently only possible by doing fuzzy back of the napkin math on the lambda's concurrent executions graph. This is problematic when trying to figure out definitively when shared state in the process has been created or released. For example, we have connections in a shared HTTP connection pools to services that require authenticated connections. However, we lack visibility into when our lambdas are being restarted and it's hard to definitively answer the question, "are we thrashing service xyz's authentication process right now?"

**Describe the solution you'd like**
Application log message when the express.js server is started and stopped so we can create log insight reports of starts and stops.

**Describe alternatives you've considered**
Using the Concurrent Executions metric and reporting to get a general idea.

**Additional context**
Happy to do the PR here.

ssvegaraju:
This is a good point and a useful metric; I've added this to our team's backlog. Would also be happy to review a PR if you already have one in the works or plan to work on it. Thanks for bringing this up!

Zambonilli:
PR created, https://github.com/awslabs/fhir-works-on-aws-deployment/pull/627

nisankep:
FHIR Works on AWS has been moved to maintenance mode. While in maintenance, we will not add any new features to this solution. All security issues should be reported directly to AWS Security at [aws-security@amazon.com] (mailto:security@amazon.com). If you are new to this solution, we advise you to explore using [HealthLake] (https://aws.amazon.com/healthlake), which is our managed service for building FHIR based transactional and analytics applications. You can get started by contacting your AWS Account team. If you are an existing customer
of FHIR Works on AWS, and have additional questions or need immediate help, please reach out to [fwoa-migration-support@amazon.com](mailto:fwoa-migration-support@amazon.com) or contact your AWS Account team.

PR #627, opened 2022-05-17
feat: ISSUE 624 add logging when lambda starts and stops fwoa applica…

Zambonilli:
…tion code in the fhirServer lambda

Issue #, if available:
https://github.com/awslabs/fhir-works-on-aws-deployment/issues/624

Description of changes:
Added logging for when the fhirServer lambda process is started or stopped. I was able to test by running `serverless offline start --printOutput --allowCache --logLevel info`

I was able to assert the following:
* the 1st web request processed writes `start` entry
* subsequent requests do not log a process event
* kill -15 $pid writes `SIGTERM` entry & `exit` entry
* kill -2 $pid writes `SIGINT` entry & `exit` entry

Checklist:
* [X ] Have you successfully deployed to an AWS account with your changes?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,8 +4,21 @@
  */
 
 import serverless from 'serverless-http';
+import { partial } from 'lodash';
 import { generateServerlessRouter } from 'fhir-works-on-aws-routing';
+import { makeLogger } from 'fhir-works-on-aws-interface';
 import { getFhirConfig, genericResources } from './config';
+
+// setup logging for process start and stops
+const logger = makeLogger({ pid: process.pid, ppid: process.ppid });
+const logProcessEvent = (eventName: string, codeOrSignal: any) => {
+    logger.info('process event raised', { eventName, codeOrSignal });
+};
+logProcessEvent('start', undefined);
+process.on('beforeExit', partial(logProcessEvent, 'beforeExit'));
+process.on('exit', partial(logProcessEvent, 'exit'));
+process.on('SIGTERM', partial(logProcessEvent, 'SIGTERM'));
+process.on('SIGINT', partial(logProcessEvent, 'SIGINT'));
 
 const ensureAsyncInit = async (initPromise: Promise<any>): Promise<void> => {
     try {
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,8 +11,15 @@ import { getFhirConfig, genericResources } from './config';
 
 // setup logging for process start and stops
 const logger = makeLogger({ pid: process.pid, ppid: process.ppid });
-const logProcessEvent = (eventName: string, codeOrSignal: any) => {
+const logProcessEvent = async (eventName: string, codeOrSignal: any) => {
     logger.info('process event raised', { eventName, codeOrSignal });
+
+    if (eventName === 'SIGTERM') {
+        // console.* methods are async when the process is piped
+        // so sleep 200ms out of the 300ms allocated to SIGTERM
+        // https://nodejs.org/api/process.html#a-note-on-process-io
+        await new Promise((resolve) => setTimeout(resolve, 200));
+    }
 };
 logProcessEvent('start', undefined);
 process.on('beforeExit', partial(logProcessEvent, 'beforeExit'));


Bingjiling:
Hi @Zambonilli,

Thanks for your contribution! Were you able to test this functionality in an AWS account? If so, could you please share a screenshot of what the generated CloudWatch log looks like? 

Thanks,
Yanyu

Zambonilli:
Yes, we've deployed this PR to an AWS account. We're seeing the `start` process event logged correctly. However, the `exit`, `beforeExit`, `SIGINT` and `SIGTERM` are not being logging in AWS environments. We believe this is due to firecracker hard killing the process with a kill -9 or not waiting for the nodejs process to complete all outstanding event loop I/O, in our case the buffering of our console writes to std out, before killing the process.

Screen Shot 2022-05-23 at 12 00 07 PM

Zambonilli:
Looking more into the behavior of the various console methods when exiting. I believe the reason we're not seeing synchronous behavior is because the lambda runtime is piping the std out and std error for the process. According to the nodejs docs, piping behavior makes the console methods asynchronous. https://nodejs.org/api/process.html#a-note-on-process-io

Looking at the aws-samples for graceful shutdown, the nodejs example does an artificial sleep and indicates that you can replace with your shutdown async I/O for the same effect. The sleep used here is 200ms which is 67% of the default allocated SIGTERM time of 300ms before lambda executes a SIGKILL on the process. https://github.com/aws-samples/graceful-shutdown-with-aws-lambda/blob/main/nodejs-demo/hello-world/app.js#L8

So, I'm going to add an artificial sleep here of 200ms to the function used to log process events. While this logic is non-deterministic, it theoretically will give some amount of time for the process to attempt to asynchronously write the messages to std out.

Zambonilli:
Deployed the version with the 200ms delay to an AWS environment and did not see the SIGTERM console events in the cloudwatch logs. I'm at a loss as to why the exit events are not writing to the logs here. The root cause looks to be the way in which lambda is killing the process and does not look to match the behavior documented in the aws-sample for graceful shutdown or any of the stackoverflow threads on the subject.

nguyen102:
@Zambonilli doesn't lambda already emit metrics for a lambda's duration? Can you use those metrics to get that datapoint instead of using CWL?
https://docs.aws.amazon.com/lambda/latest/dg/monitoring-metrics.html

Zambonilli:
Unfortunately, I don't think so. What we're mainly interested in here is some sort of metrics on the shared state of running lambda instances and not just invocations. Because shared state is tied to the provisioning and deprovisioning of the lambda instances themselves and not necessarily invocations.

We hit a scenario where a SaaS service was rate limiting our tenant and our instance of fwoa was a client of the SaaS service. We were unable to even tell how many open connections to the SaaS service we had at any point during the rate limiting period because our connection pool was created at lambda process start. The best we could come up with is to generally correlate concurrent executions but that doesn't account for provisioned and not executing lambdas.

nguyen102:
Hi @Zambonilli ,

Thanks for adding these logging statements. However, I'm hesitant to add this PR to `mainline` since I don't think most users will find these logs useful. The logs are useful for debugging the performance issue you're currently facing, but I don't think they should always be kept on.

nguyen102:
Hi @Zambonilli ,

Thanks for adding these logging statements. However, I'm hesitant to add this PR to `mainline` since I don't think most users will find these logs useful. The logs are useful for debugging the performance issue you're currently facing, but I don't think they should always be kept on.

Zambonilli:
The reason I went down the path to adding these logs as opposed to metrics around our connection pool for our SaaS provider was because they are generic application log entries for process events that can be used to triage any issues around cold boot, exit and process management in lambda. I find it hard to believe that others running fwoa in production, wouldn't have scenarios where they wouldn't want these process events to triage production issues.

nguyen102:
I see. Let me discuss this further with the team. In the meanwhile, you mentioned that the `start` event was logged correctly, but not the other events. Would you be ok with removing the other logged events from the PR until we can verify that those events are logged properly?

Zambonilli:
Thanks and yes, currently we're only seeing the `start` event written to the application logs in lambda. However, we are seeing the other events logged locally when using serverless offline.

I sifted through the firecracker code a bit to see if I could spot something that would explain not logging on exit. However, it looks like firecracker just passes through the signal and cleans up the microvm resources. So, I believe the 300ms SIGTERM -> SIGKILL behavior described in the aws-samples/graceful-shutdown-with-aws-lambda was the internal lambda orchestration around firecracker and changed since the aws-samples was published. I can remove the logging for the various exit events but we'd lose that logging if AWS lambda changed their internal orchestration logic around safe exiting.

https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/signal_handler.rs#L155

nguyen102:
I'm confuse by this sentence.
>  I can remove the logging for the various exit events **but we'd lose that logging if AWS lambda changed their internal orchestration logic around safe exiting**.

Since logging the exit events doesn't currently work, we wouldn't be losing anything by removing them. We should remove all events that can not be logged correctly on AWS. I've discussed with the team and we're currently considering the option of having this log be a configuration option. The default config option would be to turn the log off. Would you be ok with this approach?

Zambonilli:
Logging the exit events works when the process is shutdown gracefully. For whatever reason, AWS lambda internally is not gracefully shutting down the process so we can't even log that we're shutting down. What I'm saying is that at some point AWS lambda internally was signalling to gracefully shutdown the process and could again at any point and we'd automatically begin to log shutdowns once lambda starts signalling graceful shutdowns.

I don't think this should be a config value. It's completely normal for software products to log to the console at the beginning of it's boot and exit processes. I'll pull the PR.

PR #637, opened 2022-06-03
feat: add implementation guides support via cdk

ssvegaraju:
Issue #, if available:

Description of changes:
Added a context parameter to specify fhir version and made it part of the main stack, so we don't have to go into the `javaHapiValidatorLambda` directory to deploy. (will document changes in the documentation task)

I tested this by running the integration test for implementation guides and ensuring that all tests passed.

Checklist:

### DIFF
--- a/bin/cdk-infra.ts
+++ b/bin/cdk-infra.ts
@@ -10,6 +10,7 @@ import FhirWorksStack from '../lib/cdk-infra-stack';
 const app = new cdk.App();
 
 const allowedLogLevels = ['error', 'info', 'debug', 'warn'];
+const allowedFHIRVersions = ['4.0.1', '3.0.1']
 
 const region: string = app.node.tryGetContext('region') || 'us-west-2';
 const stage: string = app.node.tryGetContext('stage') || 'dev';
@@ -17,9 +18,17 @@ const enableMultiTenancy: boolean = app.node.tryGetContext('enableMultiTenancy')
 const enableSubscriptions: boolean = app.node.tryGetContext('enableSubscriptions') || false;
 const oauthRedirect: string = app.node.tryGetContext('oauthRedirect') || 'http://localhost';
 const useHapiValidator: boolean = app.node.tryGetContext('useHapiValidator') || false;
-let logLevel: string = app.node.tryGetContext('logLevel') || 'error';
 const enableESHardDelete: boolean = app.node.tryGetContext('enableESHardDelete') || false;
 const enableBackup: boolean = app.node.tryGetContext('enableBackup') || false;
+let logLevel: string = app.node.tryGetContext('logLevel') || 'error';
+let fhirVersion: string = app.node.tryGetContext('fhirVersion') || '4.0.1';
+
+if (useHapiValidator) {
+    if (!allowedFHIRVersions.includes(fhirVersion)) {
+        console.log(`invalid FHIR Version specified: ${fhirVersion}`);
+        fhirVersion = '4.0.1';
+    }
+}
 
 if (!allowedLogLevels.includes(logLevel)) {
     console.log(`invalid log level specified: ${logLevel}`);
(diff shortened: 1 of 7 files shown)


bin/cdk-infra.ts
```diff
@@ -10,16 +10,25 @@ import FhirWorksStack from '../lib/cdk-infra-stack';
...
-let logLevel: string = app.node.tryGetContext('logLevel') || 'error';
 const enableESHardDelete: boolean = app.node.tryGetContext('enableESHardDelete') || false;
 const enableBackup: boolean = app.node.tryGetContext('enableBackup') || false;
+let logLevel: string = app.node.tryGetContext('logLevel') || 'error';
+let fhirVersion: string = app.node.tryGetContext('fhirVersion') || '4.0.1';
+
+if (useHapiValidator) {
+    if (!allowedFHIRVersions.includes(fhirVersion)) {
+        console.log(`invalid FHIR Version specified: ${fhirVersion}`);
+        fhirVersion = '4.0.1';
```
Bingjiling: Should we fail the deployment directly if the fhir version specified is not allowed?

lib/cdk-infra-stack.ts
```diff
@@ -601,18 +637,17 @@ export default class FhirWorksStack extends Stack {
             }),
             tracing: Tracing.ACTIVE,
         });
-        // pending port of hapi validator stack
```
Bingjiling, line 604: Remove it if it's not needed?

lib/cdk-infra-stack.ts
```diff
@@ -501,14 +514,37 @@ export default class FhirWorksStack extends Stack {
...
+            currentVersionOptions: {
+                provisionedConcurrentExecutions: 5,
+            },
             environment: {
                 ...lambdaDefaultEnvVars,
                 EXPORT_STATE_MACHINE_ARN: bulkExportStateMachine.bulkExportStateMachine.stateMachineArn,
                 PATIENT_COMPARTMENT_V3,
                 PATIENT_COMPARTMENT_V4,
+                // config.ts only checks if the arn is present or equal to [object Object]
+                VALIDATOR_LAMBDA_ALIAS: props!.useHapiValidator ? this.javaHapiValidator!.hapiValidatorLambda.functionArn : '[object Object]',
```
Bingjiling: Is there a way to update the config.ts check so there's no need to assign '[object Object]' to ALIAS value? It looks quite hacky.

lib/cdk-infra-stack.ts
```diff
@@ -501,14 +514,37 @@ export default class FhirWorksStack extends Stack {
...
+                    beforeBundling(inputDir, outputDir) {
+                        return [];
+                    },
+                    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                    beforeInstall(inputDir, outputDir) {
+                        return [];
+                    },
+                    afterBundling(inputDir, outputDir) {
+                        // copy all the necessary files for the lambda into the bundle
+                        // this allows the lambda functions for bulk export to have access to these files within the lambda instance
```
Bingjiling: Update the comment for implementation guide

Bingjiling (commented):
Looks good! A few comments -

bin/cdk-infra.ts
ssvegaraju: This makes more sense, changing this to throw an error instead.

lib/cdk-infra-stack.ts
ssvegaraju: If this is referring to the comment, it's already been removed.

lib/cdk-infra-stack.ts
ssvegaraju: I changed it to also check for empty strings so we could use that instead. Unfortunately, i wasn't able to use `undefined`, so i had to use a string of some sort since that is the type it was expecting.

lib/cdk-infra-stack.ts
ssvegaraju: Thanks for catching this- updated.

lib/cdk-infra-stack.ts
Bingjiling: What about the commented out section below? Is it needed?

lib/cdk-infra-stack.ts
Bingjiling: Empty string works! Thanks for diving in this one.

Bingjiling (approved):

PR #656, opened 2022-07-19
fix: update paths and include linux esbuild

ssvegaraju:
Issue #, if available:

Description of changes:
Updating paths to use forward slashes instead of backslashes as per https://stackoverflow.com/a/38428899. In addition, created a .yarnrc.yml to allow for multiple versions of esbuild based on platform.

Tested using WSL.

Checklist:

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,4 +1,0 @@
-supportedArchitectures:
-  os:
-    - "linux"
-    - "win64"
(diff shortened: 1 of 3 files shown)


.yarnrc.yml
```diff
@@ -0,0 +1,4 @@
+supportedArchitectures:
```
rsmayda: i htink we can remove this file?

rsmayda (approved):

PR #697, opened 2022-09-14
test: add search scope-based tests

ssvegaraju:
Issue #, if available:

Description of changes:
Adding some tests for SMART scopes and searches. Tested by running yarn int-test locally.
image

Checklist:

### DIFF
--- a/integration-tests/search.test.ts
+++ b/integration-tests/search.test.ts
@@ -527,55 +527,76 @@ describe('searches with patient permission levels', () => {
             class: {
                 system: 'http://terminology.hl7.org/CodeSystem/v3-ActCode',
                 code: 'IMP',
-                display: 'inpatient encounter'
+                display: 'inpatient encounter',
             },
             subject: {
-              reference: `Patient/92e0d921-bb19-4cae-a3cc-9d3c5bcf7a39`,
-              display: 'Sherlock Holmes'
-            }
+                reference: `Patient/92e0d921-bb19-4cae-a3cc-9d3c5bcf7a39`,
+                display: 'Sherlock Holmes',
+            },
         };
         encounterResource = (await adminClient.post('Encounter', encounterResourceJson)).data;
         await waitForResourceToBeSearchable(adminClient, encounterResource);
     });
-    
-    afterAll(async() => {
+
+    afterAll(async () => {
         await adminClient.delete(`Encounter/${encounterResource.id}`);
     });
 
     test('tests with only patient.read scope', async () => {
-        let client = await getFhirClient('fhirUser patient/Patient.read launch/patient profile openid', false);
-        let sherlockHolmes = (await adminClient.get(`Patient/${idsOfFhirResources.sherlockHolmes}`)).data;
+        const client = await getFhirClient('fhirUser patient/Patient.read launch/patient profile openid', false);
+        const sherlockHolmes = (await adminClient.get(`Patient/${idsOfFhirResources.sherlockHolmes}`)).data;
 
-        
-        let getPatient = await client.get('Patient');
+        const getPatient = await client.get('Patient');
         expect(getPatient.data.total).toBe(2); // Should return Mycroft Holmes as a seealso
         expect(getPatient.data.entry[1].resource).toMatchObject(sherlockHolmes);
 
-        await expectResourceToNotBePartOfSearchResults(client, {url: 'Patient', params: {_include: 'Patient:organization'}}, {
-            resourceType: 'Organization',
-        });
+        await expectResourceToNotBePartOfSearchResults(
+            client,
+            { url: 'Patient', params: { _include: 'Patient:organization' } },
+            {
+                resourceType: 'Organization',
+            },
+        );
 
-        await expectResourceToNotBePartOfSearchResults(client, { url: 'Patient', params: {_revinclude: 'Encounter:subject'}}, encounterResource);
+        await expectResourceToNotBePartOfSearchResults(
+            client,
+            { url: 'Patient', params: { _revinclude: 'Encounter:subject' } },
+            encounterResource,
+        );
 
         // failing searches (no scopes for Observation, so should reject)
         await expect(client.get('Observation')).rejects.toMatchObject({
             response: { status: 401 },
         });
-        
+
         await expect(client.get('Observation', { params: { _include: 'Observation:subject' } })).rejects.toMatchObject({
             response: { status: 401 },
         });
     });
 
     test('tests with additional patient scopes', async () => {
-        let orgClient = await getFhirClient('fhirUser patient/Patient.read patient/Organization.read launch/patient profile openid', false);
-        let encClient = await getFhirClient('fhirUser patient/Patient.read patient/Encounter.read launch/patient profile openid', false);
+        const orgClient = await getFhirClient(
+            'fhirUser patient/Patient.read patient/Organization.read launch/patient profile openid',
+            false,
+        );
+        const encClient = await getFhirClient(
+            'fhirUser patient/Patient.read patient/Encounter.read launch/patient profile openid',
+            false,
+        );
 
-        await expectResourceToNotBePartOfSearchResults(orgClient, {url: 'Patient', params: {_include: 'Patient:organization'}}, {
-            resourceType: 'Organization',
-        });
+        await expectResourceToNotBePartOfSearchResults(
+            orgClient,
+            { url: 'Patient', params: { _include: 'Patient:organization' } },
+            {
+                resourceType: 'Organization',
+            },
+        );
 
-        await expectResourceToBePartOfSearchResults(encClient, { url: 'Patient', params: {_revinclude: 'Encounter:subject'}}, encounterResource);
+        await expectResourceToBePartOfSearchResults(
+            encClient,
+            { url: 'Patient', params: { _revinclude: 'Encounter:subject' } },
+            encounterResource,
+        );
     });
 });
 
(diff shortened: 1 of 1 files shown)


integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        };
+        encounterResource = (await adminClient.post('Encounter', encounterResourceJson)).data;
+        await waitForResourceToBeSearchable(adminClient, encounterResource);
+    });
+
+    afterAll(async () => {
+        await adminClient.delete(`Encounter/${encounterResource.id}`);
+    });
+
+    test('tests with only patient.read scope', async () => {
```
rsmayda:
```suggestion
    test('tests with only user/Patient.read scope', async () => {
```

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        });
+
+        await expect(
+            clientAsAdmin.get('Observation', { params: { _include: 'Observation:subject' } }),
+        ).rejects.toMatchObject({
+            response: { status: 401 },
+        });
+    });
+
+    test('tests with additional patient scopes', async () => {
```
rsmayda:
```suggestion
    test('tests with additional user scopes', async () => {
```

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
         });
     });
 });
+
+describe('searches with patient permission levels', () => {
```
rsmayda:
```suggestion
describe('searches with patient scopes', () => {
```

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        };
+        encounterResource = (await adminClient.post('Encounter', encounterResourceJson)).data;
+        await waitForResourceToBeSearchable(adminClient, encounterResource);
+    });
+
+    afterAll(async () => {
+        await adminClient.delete(`Encounter/${encounterResource.id}`);
+    });
+
+    test('tests with only patient.read scope', async () => {
```
rsmayda:
```suggestion
    test('tests with only patient/Patient.read scope', async () => {
```

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        const getPatient = await client.get('Patient');
+        expect(getPatient.data.total).toBe(2); // Should return Mycroft Holmes as a seealso
+        expect(getPatient.data.entry[1].resource).toMatchObject(sherlockHolmes);
+
+        await expectResourceToNotBePartOfSearchResults(
+            client,
+            { url: 'Patient', params: { _include: 'Patient:organization' } },
+            {
+                resourceType: 'Organization',
+            },
```
rsmayda: Is this detailed enough to catch if Org is included or not? For the `user` test is that the case?

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        const orgClient = await getFhirClient(
+            'fhirUser patient/Patient.read patient/Organization.read launch/patient profile openid',
+            false,
+        );
+        const encClient = await getFhirClient(
+            'fhirUser patient/Patient.read patient/Encounter.read launch/patient profile openid',
+            false,
+        );
+
+        await expectResourceToNotBePartOfSearchResults(
```
rsmayda: Add a quick comment as to why this should not be in search result -- mainly that the patient has scope access to the Organization but not resource access

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        await expectResourceToBePartOfSearchResults(
+            encClient,
+            { url: 'Patient', params: { _revinclude: 'Encounter:subject' } },
+            encounterResource,
+        );
+    });
+});
+
+// these are largely similar to above, but have different behavior and so are kept separate
+describe('searches with user permission levels', () => {
```
rsmayda:
```suggestion
describe('searches with user scope as a Practitioner (admin)', () => {
```

integration-tests/search.test.ts
rsmayda: With this set we also need to add the `user` scope as Patient;; this would be the same results/tests so feel free to use cases like here https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/src/smartAuthorizationHelper.test.ts#L214-L215

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+        encounterResource = (await adminClient.post('Encounter', encounterResourceJson)).data;
+        await waitForResourceToBeSearchable(adminClient, encounterResource);
+    });
+
+    afterAll(async () => {
+        await adminClient.delete(`Encounter/${encounterResource.id}`);
+    });
+
+    test('tests with only patient.read scope', async () => {
+        const clientAsPatient = await getFhirClient('fhirUser user/Patient.read profile openid', false);
```
rsmayda: I would make this its own `describe` or you can do the `case` example I gave above

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+                display: 'Sherlock Holmes',
+            },
+        };
+        encounterResource = (await adminClient.post('Encounter', encounterResourceJson)).data;
+        await waitForResourceToBeSearchable(adminClient, encounterResource);
+    });
+
+    afterAll(async () => {
+        await adminClient.delete(`Encounter/${encounterResource.id}`);
+    });
```
rsmayda: you can nest this at a higher level so that we only do this once across all tests
```
describe
  before..
  after...
  <existing describe tests>
```

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+            resourceType: 'Organization',
+            name: 'Health Level Seven International',
+            alias: ['HL7 International'],
+        };
+        const orgResource = (await adminClient.post('Organization', orgJson)).data;
+        const randPatient = randomPatient();
+        randPatient.managingOrganization.reference = `Organization/${orgResource.id}`;
+        const testPatient: ReturnType<typeof randomPatient> = (await adminClient.post('Patient', randPatient)).data;
+        // wait for the patient to be asynchronously written to ES
+        await waitForResourceToBeSearchable(adminClient, testPatient);
```
rsmayda: Could we move this all to the beforeall section?

integration-tests/search.test.ts
```diff
@@ -513,3 +514,242 @@ describe('search', () => {
...
+
+        await expectResourceToBePartOfSearchResults(
+            encClientAdmin,
+            { url: 'Patient', params: { name: 'Sherlock', _revinclude: 'Encounter:subject' } },
+            encounterResource,
+        );
+
+        // cleanup
+        await adminClient.delete(`Organization/${orgResource.id}`);
+        await adminClient.delete(`Patient/${testPatient.id}`);
```
rsmayda: Could move to after all section?

integration-tests/search.test.ts
ssvegaraju: I'll use the suggestions from below to create the org resource beforeAll and use that to check instead!

integration-tests/search.test.ts
```diff
@@ -536,220 +539,148 @@ describe('searches with patient permission levels', () => {
...
+            resourceType: 'Organization',
+            name: 'Health Level Seven International',
+            alias: ['HL7 International'],
+        };
+        orgResource = (await adminClient.post('Organization', orgJson)).data;
+        const randPatient = randomPatient();
+        randPatient.managingOrganization.reference = `Organization/${orgResource.id}`;
+        testPatient = (await adminClient.post('Patient', randPatient)).data;
+        // wait for the patient to be asynchronously written to ES
+        await waitForResourceToBeSearchable(adminClient, testPatient);
```
rsmayda, line 554: Do we need any other waits? or would we be okay

integration-tests/search.test.ts
```diff
@@ -536,220 +539,148 @@ describe('searches with patient permission levels', () => {
...
-
-        await expectResourceToNotBePartOfSearchResults(
-            client,
-            { url: 'Patient', params: { _revinclude: 'Encounter:subject' } },
-            encounterResource,
-        );
+    // Since the tests for patient scope and user (as Patient) scope are the same results,
+    // we can group them
+    const scopes: string[] = ['patient', 'user'];
+    describe.each(scopes)('searches with patient/user (as Patient) scopes', (scope) => {
```
rsmayda:
```suggestion
    describe.each(scopes)('searches with %p (as Patient) scopes', (scope) => {
```

This should name the test correctly

integration-tests/search.test.ts
```diff
@@ -515,9 +515,12 @@ describe('search', () => {
     });
 });
 
-describe('searches with patient permission levels', () => {
+describe('test searches with different scopes', () => {
     let adminClient: AxiosInstance;
```
rsmayda, line 519: This can be moved to the `beforeAll` context and made `const`

ssvegaraju: It's also used in `afterAll()`, which is why I kept it out

integration-tests/search.test.ts
ssvegaraju: This should be waiting for everything correctly!

integration-tests/search.test.ts
rsmayda: ops thanks! please ignore

integration-tests/search.test.ts
```diff
@@ -513,3 +514,173 @@ describe('search', () => {
...
+    afterAll(async () => {
+        await adminClient.delete(`Encounter/${encounterResource.id}`);
+        await adminClient.delete(`Organization/${orgResource.id}`);
+        await adminClient.delete(`Patient/${testPatient.id}`);
+    });
+
+    // Since the tests for patient scope and user (as Patient) scope are the same results,
+    // we can group them
+    const scopes: string[] = ['patient', 'user'];
+    describe.each(scopes)('searches with %p scopes', (scope) => {
```
rsmayda:
```suggestion
    describe.each(scopes)('search with %p scope type as a patient', (scope) => {
```

rsmayda (approved):
Looks good thank you for your hard work

Issue #671, opened 2022-08-04
[Bug] Unable to compile US Core Implementation Guides 


yonghaochen-nt:
**Describe the bug**
I followed the [Using FHIR Implementation Guides](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/USING_IMPLEMENTATION_GUIDES.md#using-fhir-implementation-guides), created `implementationGuides` directory and downloaded us-core IG packages and dependencies.  When I use `yarn run compile-igs` in the parent directory, it throw the error below:

Exception:  The following input is not a search parameter: {
  "code": "end",
  "description": "Optional Extensions Element",
  "experimental": true,
  "id": "valueset-extensions-ValueSet-end",
  "name": "end",
  "resourceType": "SearchParameter",
  "status": "draft",
  "type": "date",
  "url": "[http://hl7.org/fhir/SearchParameter/valueset-extensions-ValueSet-end"](http://hl7.org/fhir/SearchParameter/valueset-extensions-ValueSet-end%22),
  "version": "4.0.1",
  "xpath": "f:ValueSet/f:extension[@url='[http://hl7.org/fhir/StructureDefinition/valueset-expirationDate']](http://hl7.org/fhir/StructureDefinition/valueset-expirationDate%27]) | /f:#expirationDate",
  "xpathUsage": "normal"
}

**Expected behavior**
the us core IG and its dependencies could be compiled without errors.

ssvegaraju:
Hi @yonghaochen-nt, could you provide a link to the us-core IG and dependencies you are trying to compile? Also, if you could list the steps you took when adding them to the `implementationGuides` directory, that would be very helpful for us to try and reproduce this bug on our end!

Thanks,
Sukeerth

yonghaochen-nt:
> Hi @yonghaochen-nt, could you provide a link to the us-core IG and dependencies you are trying to compile? Also, if you could list the steps you took when adding them to the `implementationGuides` directory, that would be very helpful for us to try and reproduce this bug on our end!
> 
> Thanks, Sukeerth

Hi @ssvegaraju , us-core IG is from [https://build.fhir.org/ig/HL7/US-Core/package.tgz](url). Dependencies are 
[hl7.fhir.r4.examples](https://simplifier.net/packages/hl7.fhir.r4.examples/4.0.1)
[hl7.fhir.uv.bulkdata](https://simplifier.net/packages/hl7.fhir.uv.bulkdata)
[...]

Issue #642, opened 2022-06-27
[Bug] IGs since IG Publisher 1.1.119 fail to parse


Zambonilli:
**Describe the bug**
We're seeing our Implementation Guide fail to be parsed by scripts/compile-igs.ts with `Exception:  Missing dependency hl7.terminology.r4@3.1.0 Error: Missing dependency hl7.terminology.r4@3.1.0`. We have not created a custom terminology server and are not using any other IGs. We've been able to identify that as of IG publisher version 1.1.119 a new dependency is added to the IG tgz's package.json for `hl7.terminology.r4` : `3.1.0`. Rolling back to 1.1.118 IG publisher fixes the fwoa compiler error and the script completes successfully. We haven't had enough time to root out exactly what the issue is yet and there's nothing that jumps out in the IG publisher 1.1.119 release notes to indicate the root cause. We do however believe that since our IG is so basic that other people creating IGs will have the same issue in fwoa compiler.

IG Publisher 1.1.118 dependencies:
```
"dependencies": {
    "hl7.fhir.r4.core": "4.0.1"
  },
```

IG Publisher 1.1.119+ dependencies
```
"dependencies": {
    "hl7.fhir.r4.core": "4.0.1",
    "hl7.terminology.r4": "3.1.0"
  },
```

**To Reproduce**
1. Use IG Publisher 1.1.119 or greater to generate a tgz of your IG
2. Navigate to the deployment repo 
3. yarn install
4. yarn run compile-igs
5. assert that the compiler returned the error `Exception:  Missing dependency hl7.terminology.r4@3.1.0 Error: Missing dependency hl7.terminology.r4@3.1.0`

**Expected behavior**
Successfully compiled IG. Although, I'm not sure if the IG Publisher is now producing different output since it looks like there are pushes for r4b in the more recent releases. This could be an error with the IG publisher no longer being backwards compatible with r4.0.1.

**Versions (please complete the following information):**
v2.5.1-smart

**Additional context**
https://github.com/HL7/fhir-ig-publisher/releases/tag/1.1.119

SanketD92:
Thank you for creating this bug report. I'll create a ticket in our team's backlog.

ssvegaraju:
Hi,
[...]

PR #789, opened 2023-01-18
feat: implementation guides on s3

ssvegaraju:
Issue #390, #642, #671

Description of changes:
Refactored Validator to use S3 to support larger IG file sizes.

Checklist:

* [ ] Did you add new integration tests?
* [x] Did you run integration tests with your changes?


javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +59,51 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+    }
+
+    private Map<String, List<IGObject>> downloadObjects(List<String> keys, String bucketName) {
+        // keep track of .index.json objects
+        List<IGObject> indices = new ArrayList<IGObject>();
+        List<IGObject> resources = new ArrayList<IGObject>();
+        for (String key : keys) {
+            try (S3ObjectInputStream s3Object = s3.getObject(bucketName, key).getObjectContent()) {
+                IGObject bucketObj = new Utils.IGObject(key, IOUtils.toString(s3Object));
+                if (key.contains(".index.json")) {
```
sevu1, line 93: Do we want to check case insentitive?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +59,51 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+                    indices.add(bucketObj);
+                } else {
+                    resources.add(bucketObj);
+                }
+            } catch (Exception e) {
+                log.error(e.getMessage());
+                throw new Error(e);
```
sevu1: Do we want to rethrow original error?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +59,51 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+        for (String key : keys) {
+            try (S3ObjectInputStream s3Object = s3.getObject(bucketName, key).getObjectContent()) {
+                IGObject bucketObj = new Utils.IGObject(key, IOUtils.toString(s3Object));
+                if (key.contains(".index.json")) {
+                    indices.add(bucketObj);
+                } else {
+                    resources.add(bucketObj);
+                }
+            } catch (Exception e) {
+                log.error(e.getMessage());
```
sevu1: Should we log entire error object instead of just message?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: The file is named `.index.json` by standard so I don't think it's necessary

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: updated to log the entire object!

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: Do you mean directly throwing the Exception? Is that preferable to wrapping it in an Error?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
```diff
@@ -38,4 +59,51 @@ public ValidatorResponse handleRequest(String event, Context context) {
...
+        }
+    }
+
+    private Map<String, List<IGObject>> downloadObjects(List<String> keys, String bucketName) {
+        // keep track of .index.json objects
+        List<IGObject> indices = new ArrayList<IGObject>();
+        List<IGObject> resources = new ArrayList<IGObject>();
+        for (String key : keys) {
+            try (S3ObjectInputStream s3Object = s3.getObject(bucketName, key).getObjectContent()) {
+                IGObject bucketObj = new Utils.IGObject(key, IOUtils.toString(s3Object));
```
Bingjiling: So the validator do not need to care how many IGs are included, if there're two IGs to load, there'll be two .index.json file and related resource files do not need to be differentiated between two IGs?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Utils.java
```diff
@@ -0,0 +1,28 @@
...
+        public String getKey() {
+            return key;
+        }
+
+        public String getContent() {
+            return content;
+        }
+    }
+
+    public static List<IGObject> EmptyIfNull(List<IGObject> iterable) {
```
Bingjiling: Where is this function used?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Utils.java
```diff
@@ -0,0 +1,28 @@
+package software.amazon.fwoa;
+
+import java.util.Collections;
+import java.util.List;
+
+public final class Utils {
```
Bingjiling: Nit: this class can have a more specific name like IGUtils?

lib/javaHapiValidator.ts
```diff
@@ -18,10 +23,47 @@ export default class JavaHapiValidator extends Stack {
...
+            enableKeyRotation: true,
+            description: 'KMS CMK for IGs',
+            policy: new PolicyDocument({
+                statements: [
+                    new PolicyStatement({
+                        sid: 'Enable IAM Root Permissions',
+                        effect: Effect.ALLOW,
+                        actions: ['kms:*'],
+                        resources: ['*'],
+                        principals: [new AnyPrincipal()],
```
Bingjiling: Should this be `AccountRootPrincipal()`?

USING_IMPLEMENTATION_GUIDES.md
```diff
@@ -32,14 +32,22 @@ The prerequisites for FHIR IGs are same as in the FHIR [installation documentati
...
+1. Deploy the Hapi Validator using the following commands for CDK:
 
   When deploying, simply append the useHapiValidator flag (and the optional fhirVersion flag, which defaults to 4.0.1):
   ```sh
-  yarn deploy -c useHapiValidator=true -c fhirVersion=4.0.1
+  yarn deploy -c useHapiValidator=true -c fhirVersion=4.0.1 --all
   ```
 
-  Or, with Serverless, continue with these steps: (LEGACY)
+  If you are deploying with implementation guides that are large in file size such as [us.nlm.vsac](https://registry.fhir.org/package/us.nlm.vsac|0.3.0), please increase the memory values in [the CDK template](./lib/javaHapiValidator.ts). The values to consider upgrading for your use case are the `memoryLimit` of the BucketDeployment, and the `memorySize` and `ephemeralStorageSize` values for the Validator Lambda Function.
```
Bingjiling: Can we set these values as cdk context or something easier for a customer to update?

lib/javaHapiValidator.ts
```diff
@@ -34,8 +76,12 @@ export default class JavaHapiValidator extends Stack {
             tracing: Tracing.ACTIVE,
             environment: {
                 FHIR_VERSION: props.fhirVersion,
+                IMPLEMENTATION_GUIDES_BUCKET: igDeployment.deployedBucket.bucketName,
             },
         });
         this.alias = this.hapiValidatorLambda.currentVersion.addAlias(`fhir-service-validator-lambda-${props.stage}`);
+
+        igDeployment.deployedBucket.grantReadWrite(this.hapiValidatorLambda);
```
Bingjiling: Why does the Lambda need write and delete?

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
ssvegaraju: Yup each IG is placed into its own folder and will be indexed by the .index.json file.

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Utils.java
ssvegaraju: This is used in the Validator.java file, lines 159 and 165

lib/javaHapiValidator.ts
ssvegaraju: This sadly doesn't work with the AccountRootPrincipal, no matter what permission set I tried. Runs into an error when trying to decrypt due to access denied.

lib/javaHapiValidator.ts
ssvegaraju: It actually doesn't, just finished testing without them. Removed!

lib/javaHapiValidator.ts
ssvegaraju: Just kidding, I had just missed a permission on the lambda itself. Updated!

Bingjiling (approved):

javaHapiValidatorLambda/src/main/java/software/amazon/fwoa/Handler.java
sevu1: This is preferred if you want to preserve stack: https://stackoverflow.com/a/1097539 . Not sure if that's necessary as we log the error though

ssvegaraju: Gotcha, good to know!

PR #802, opened 2023-01-25
chore: cfn nag 

kcadette:
Issue #, if available:

Description of changes:

Checklist:

* [ ] Did you add new integration tests?
* [ ] Did you run integration tests with your changes?


bin/cdk-infra.ts
```diff
@@ -71,7 +71,36 @@ const stack = new FhirWorksStack(app, `smart-fhir-service-${stage}`, {
...
+    {
+        id: 'HIPAA.Security-OpenSearchInVPCOnly',
+        reason: 'We have a guide for users that would like to deploy resources inside a VPC',
+    },
+    {
+        id: 'HIPAA.Security-S3BucketReplicationEnabled',
+        reason: 'S3 bucket replication is included as best practices in the deployment guide',
+    },
+    {
+        id: 'HIPAA.Security-LambdaConcurrency',
```
ssvegaraju, line 99: We should probably create a ticket to track where this lambda is being generated from and ensure it isn't needed in our stack.

lib/alarms.ts
```diff
@@ -49,7 +49,7 @@ export default class AlarmsResource {
             alarmDescription:
                 'Alarm when the Stream errors is more than 1 unit for 15 minutes out of the past 25 minutes. Streams do have retry logic',
             alarmName: `FhirSolution.${stage}.High.DDBToESLambdaErrorAlarm`,
-            actionsEnabled: false,
+            actionsEnabled: true,
```
ssvegaraju, line 52: Do we need to do anything else like assigning actions or is just setting it to true sufficient?

lib/cdk-infra-stack.ts
```diff
@@ -345,35 +360,196 @@ export default class FhirWorksStack extends Stack {
...
+            }),
+        );
         const startExportJobLambdaFunction = new NodejsFunction(this, 'startExportJobLambdaFunction', {
             ...defaultBulkExportLambdaProps,
             handler: 'startExportJobHandler',
+            reservedConcurrentExecutions: isDev ? 10 : 200,
+            deadLetterQueue: startExportJobLambdaFunctionDLQ,
         });
 
+        const stopExportJobLambdaFunctionDLQ = new Queue(this, 'stopExportJobLambdaFunctionDLQ', {
```
ssvegaraju, line 405: Is it possible (or even a good idea) for us to use the same DLQ for all the export lambdas?

lib/cdk-infra-stack.ts
```diff
@@ -715,7 +968,7 @@ export default class FhirWorksStack extends Stack {
             ?.addResource('.well-known')
             .addResource('smart-configuration')
             .addMethod('GET', new LambdaIntegration(fhirServerLambda), {
-                authorizationType: AuthorizationType.NONE,
+                authorizationType: AuthorizationType.CUSTOM,
```
ssvegaraju: Why are these endpoint authorizationTypes being changed to CUSTOM?

lib/alarms.ts
kcadette: There was already actions added https://github.com/awslabs/fhir-works-on-aws-deployment/blob/faf2a5eecf7e78fac6f90c1f443c0c6bb7c76564/lib/alarms.ts#L68 , but they were never enabled

lib/cdk-infra-stack.ts
kcadette: cfn-nag warns that they shouldnt be NONE. But in our case it should still be None because we use a third party IdP

lib/alarms.ts
ssvegaraju: Oops 😅

lib/cdk-infra-stack.ts
kcadette: created!

lib/cdk-infra-stack.ts
kcadette: created!

ssvegaraju (approved):

